PR #11, opened 2020-01-31
[SQL-Gen] Isolate whitespace interpolation to one function

jaylees14:
Rather than adding white space in each individual generator, make them return lists of items and perform the interpolation in one single place for all queries. 

Should hopefully make it easier to add unit tests without depending too much on the formatting


Lewiky (approved):

Lewiky (approved):

c-harding (approved):
Very clean

PR #12, opened 2020-01-31
[SQL-Gen] Implement INSERT statements and WHERE clauses

Lewiky:
Implement INSERT statements in the format described [here](https://www.calhoun.io/inserting-records-into-a-postgresql-database-with-gos-database-sql-package/), using $n to represent values for prepared statements.

Also expand INSERT and SELECT statements to support the WHERE modifier.

Implemented lots of tests for all of this.

[](https://media.giphy.com/media/j3t6W63eKTYa1jMsa3/giphy.gif)


Lewiky:
Hold on this PR until some reworking of the logic is completed

c-harding (changes requested):
- There can be at most one WHERE clause
- The order of JOIN and WHERE is fixed
- WHERE should support conjunction (and disjunction?)
- INSERT doesn't have WHERE

src/main/scala/temple/generate/database/PostgresGenerator.scala
```diff
@@ -61,8 +72,16 @@ object PostgresGenerator extends DatabaseGenerator {
             .mkString(",\n")
             .pipe(indent(_))
         s"CREATE TABLE $tableName (\n$stringColumns\n);"
-      case Read(tableName, columns) =>
+      case Read(tableName, columns, condition) =>
+        val stringColumns = columns.map(_.name).mkString(", ")
+        val stringCondition = condition.map(generateCondition) match {
+          case Some(conditionString) => List(s"WHERE $conditionString")
```
jaylees14, line 80: Why is this a list? Surely we just want to produce the empty string if none?

src/main/scala/temple/generate/database/PostgresGenerator.scala
```diff
@@ -61,8 +72,16 @@ object PostgresGenerator extends DatabaseGenerator {
...
+          case None                  => List()
+        }
+        (s"SELECT $stringColumns FROM $tableName" +: stringCondition).mkString("", " ", ";")
```
jaylees14, line 83: Can we avoid the usage of `mkString` here? Maybe take it up to the `stringCondition` variable and just make this an interpolation line, since it's quite hard to read

jaylees14 (commented):
Looks good, just a couple of things that could be improved in the `Read` case

src/main/scala/temple/generate/database/PostgresGenerator.scala
c-harding: I can see the reasoning behind this, as it determines the spacing before. The alternative would be to stick a space before the word WHERE, not sure if that makes things clearer

src/main/scala/temple/generate/database/PostgresGenerator.scala
c-harding: I’m really torn here, not sure of the most elegant way of representing this.

src/main/scala/temple/generate/database/PostgresGenerator.scala
```diff
@@ -61,8 +72,16 @@ object PostgresGenerator extends DatabaseGenerator {
...
+      case Insert(tableName, columns) =>
         val stringColumns = columns.map(_.name).mkString(", ")
-        s"SELECT $stringColumns FROM $tableName;"
+        val values        = (for (i <- 1.to(columns.length)) yield s"$$$i").toList.mkString(", ")
```
c-harding: This seems pretty convoluted, but I don’t know how it could be made nicer. Can the triple dollar be avoided? I guess the best alternatives would be `.format` or `'$' + i.toString` (we need some JS-level type coercion) 😉

I’d definitely infix the word `to`: it makes the arguments seem more equal.

Not too much of a change, but personally I think a map would be more legible than a for loop in a chain. While I don’t have a compiler to hand, `toList` doesn't seem like it should be necessary.

```scala
(1 to columns.length).map(i => "$$$i").mkString(", ")
```

src/main/scala/temple/generate/database/ast/Condition.scala
```diff
@@ -0,0 +1,11 @@
+package temple.generate.database.ast
+
+/** AST Implementation of all boolean conditions supported in Templefile */
+sealed trait Condition
+
+object Condition {
```
c-harding, line 6: Do we want to add `IsNull(column)`? Note that `Inverse(IsNull(column))` is usually rendered as `column IS NOT NULL`.

src/main/scala/temple/generate/database/PostgresGenerator.scala
```diff
@@ -33,6 +35,15 @@ object PostgresGenerator extends DatabaseGenerator {
       case References(table, colName) => s"REFERENCES $table($colName)"
     }
 
+  /** Given a query modifier, generate the type required by PostgreSQL */
+  private def generateCondition(condition: Condition): String =
+    condition match {
+      case Comparison(left, comp, right) => s"$left ${generateComparison(comp)} $right"
+      case Disjunction(left, right)      => s"${generateCondition(left)} OR ${generateCondition(right)}"
```
c-harding: Parens are definitely needed in some cases, but I’d probably suggest always including them to avoid missing edge cases.

src/main/scala/temple/generate/database/PostgresGenerator.scala
```diff
@@ -33,6 +35,15 @@ object PostgresGenerator extends DatabaseGenerator {
...
+      case Conjunction(left, right)      => s"${generateCondition(left)} AND ${generateCondition(right)}"
+      case Inverse(condition)            => s"NOT ${generateCondition(condition)}"
```
c-harding: I’d include some parens here too

src/test/scala/temple/generate/database/package.scala
```diff
@@ -63,10 +64,112 @@ package object database {
         Column("dateOfBirth"),
         Column("timeOfDay"),
         Column("expiry")
-      )
+      ),
+      None
```
c-harding: Worth making this argument optional?

src/main/scala/temple/generate/database/PostgresGenerator.scala
```diff
@@ -2,17 +2,19 @@ package temple.generate.database
...
+
 import scala.util.chaining._
 
 /** Implementation of [[DatabaseGenerator]] for generating PostgreSQL */
 object PostgresGenerator extends DatabaseGenerator {
 
   /** Given a comparison, parse it into the Postgres format */
-  private def generateComparison(comparison: Comparison) =
+  private def generateComparison(comparison: ComparisonOperator) =
     comparison match {
```
c-harding, line 18: Equals is `=` not `==`, isn’t it?

c-harding: And I think `<>` is more postgres-ic than `!=`

src/main/scala/temple/generate/database/ast/ComparisonOperator.scala
```diff
@@ -0,0 +1,13 @@
+package temple.generate.database.ast
+
+/** AST representation for boolean comparisons */
+sealed trait ComparisonOperator
+
+object ComparisonOperator {
```
c-harding, line 6: Just a thought, is it worth implementing BETWEEN? I guess not, as it’s non-binary and would need a lot of rewriting for something that can be expressed in two inequalities.

c-harding (commented):
Can I have a weekend now?

src/main/scala/temple/generate/database/PostgresGenerator.scala
Lewiky: Issue with this is that there needs to be a space between the $tablename and the WHERE, but there shouldn't be a space if there is no `stringCondition`. That's why I used the `List` and mkString, as it handles that nicely - I can't think of another nice way of doing all this without explicity adding the space at the beginning of the WHERE string

c-harding: Hmmm, feels like it should be an optional then, but then there's no mkString method

src/main/scala/temple/generate/database/ast/ComparisonOperator.scala
Lewiky: Yeah I think this would be nice for completeness but the required complexity probably isn't worth it at this point

src/main/scala/temple/generate/database/PostgresGenerator.scala
```diff
@@ -39,12 +39,12 @@ object PostgresGenerator extends DatabaseGenerator {
   private def generateCondition(condition: Condition): String =
     condition match {
       case Comparison(left, comp, right) => s"$left ${generateComparison(comp)} $right"
-      case Disjunction(left, right)      => s"${generateCondition(left)} OR ${generateCondition(right)}"
-      case Conjunction(left, right)      => s"${generateCondition(left)} AND ${generateCondition(right)}"
+      case Disjunction(left, right)      => s"(${generateCondition(left)}) OR (${generateCondition(right)})"
+      case Conjunction(left, right)      => s"(${generateCondition(left)}) AND (${generateCondition(right)})"
       case Inverse(condition) =>
```
c-harding: Can this be inlined to `Inverse(IsNull(condition))` and `Inverse(condition)`?

c-harding (approved):

jaylees14 (approved):
Looks good, nice job 👍

PR #13, opened 2020-01-31
[Config] Add code coverage via Codeconv & Scoverage

Lewiky:
Integrate code coverage reports with the CI pipeline via Codeconv and Scoverage


Lewiky:
Coverage task failed here because it required an sbt plugin that's specified in the `plugins.sbt` file, which hasn't been PRed yet....

jaylees14:
@Lewiky But it's included with this PR right? That's what gets run on the CI..?

project/plugins.sbt
```diff
@@ -0,0 +1 @@
+addSbtPlugin("org.scoverage" % "sbt-scoverage" % "1.6.1")
```
c-harding: This lack of line break at the end disturbs me, what if I want to cat the file?

c-harding:
The coverage check error looks weirdly like a syntax error

c-harding (approved):
Looks good, but is `develop` the branch for which we want to share the code cov of long-term, or would `master` be better? Also worth noting that GitHub shows `develop` when the repo is opened.

jaylees14 (approved):
LGTM. As discussed, would be worth running on merge as well as during PR. Probably only want to do this for `develop` and `master` though.

PR #14, opened 2020-01-31
[cli] Add base CLI args parser

jaylees14:
- Add base CLI args parser for `generate`
- Process generate requests by reading file and passing through parser - imagine this will change once we have some semantics attached to the AST

### DIFF
--- a/src/main/scala/temple/Main.scala
+++ b/src/main/scala/temple/Main.scala
@@ -3,11 +3,13 @@ package temple
 import temple.DSL.DSLProcessor
 import temple.utils.FileUtils
 
+/** Main entry point into the application */
 object Main extends App {
   val config = new TempleConfig(args)
   config.subcommand match {
     case Some(config.generate) => generate(config)
-    case _                     => config.printHelp()
+    case Some(_)               => throw new TempleConfig.UnhandledArgumentException
+    case None                  => config.printHelp()
   }
 
   def generate(config: TempleConfig): Unit = {
--- a/src/main/scala/temple/Main.scala
+++ b/src/main/scala/temple/Main.scala
@@ -13,10 +13,9 @@ object Main extends App {
       case None                  => config.printHelp()
     }
   } catch {
-    case error: IllegalArgumentException => {
+    case error: IllegalArgumentException =>
       System.err.println(error.getMessage)
       sys.exit(1)
-    }
   }
 
   def generate(config: TempleConfig): Unit = {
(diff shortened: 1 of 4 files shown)


Lewiky (approved):
LGTM

src/main/scala/temple/Main.scala
```diff
@@ -1,3 +1,18 @@
...
 
-object Main extends App {}
+import temple.DSL.DSLProcessor
+import temple.utils.FileUtils
+
+object Main extends App {
+  val config = new TempleConfig(args)
+  config.subcommand match {
+    case Some(config.generate) => generate(config)
+    case _                     => config.printHelp()
```
c-harding: Can we change this to None, so it’s clearer that it’s programmer error not user error if a case is forgotten here?

c-harding: And shouldn’t we print a specific error message? Or does scallop do that automatically?

jaylees14: Ah good catch with None, will change.

With regards to error message, it seems that scallop will perform `sys.exit(1)` if it encounters an error, for example:

```
sbt:temple> run foo
[info] running temple.Main foo
[scallop] Error: Excess arguments provided: 'foo'

Exception: sbt.TrapExitSecurityException thrown from the UncaughtExceptionHandler in thread "run-main-1"
[error] Nonzero exit code: 1
[error] (Compile / run) Nonzero exit code: 1
```

The case caught by `None` here is when no subcommand is provided at all: should we tell them that? The semantics I went for here mimic that of running `docker` without parameters.

c-harding: That should be fine, as long as `temple -a` errors, and it only does this on a bare input

jaylees14: Yep, that's exactly it:
```
sbt:temple> run -a
[info] running temple.Main -a
[scallop] Error: Unknown option 'a'
```

c-harding (approved):
Looks good, is it worth adding some unit tests? Check blank renders the usage, check invalid input doesn’t work, and possibly that a valid generation instruction does as expected (with mocking?)

src/main/scala/temple/TempleConfig.scala
```diff
@@ -16,3 +21,7 @@ class TempleConfig(arguments: Seq[String]) extends ScallopConf(arguments) {
   addSubcommand(generate)
   verify()
 }
+
+object TempleConfig {
+  class UnhandledArgumentException extends RuntimeException
```
Lewiky, line 26: Where is this used?

jaylees14: https://github.com/TempleEight/temple/pull/14/files#diff-ce75755896e0616e04f7efac42d83fa9R12

PR #16, opened 2020-02-03
[SQL-Gen] Add simple delete statement gen

willjvsmith:
* Add simple delete statement generation
  * Delete everything from table: `DELETE FROM Users;`
  * Delete based on condition: `DELETE FROM Users WHERE User.id = 123456;`
* Add basic unit tests

While adding this I was looking at what had changed on Friday, currently the conditions AST has Strings as their leaf node types. Do we want to add some semantics to restrict this?

`case class Comparison(left: String, comparison: ComparisonOperator, right: String) extends Condition`

In which case, which conditions do we want to support?

### DIFF
--- a/src/main/scala/temple/generate/database/PostgresGenerator.scala
+++ b/src/main/scala/temple/generate/database/PostgresGenerator.scala
@@ -85,5 +85,11 @@ object PostgresGenerator extends DatabaseGenerator {
         val stringColumns = columns.map(_.name).mkString(", ")
         val values        = (1 to columns.length).map(i => s"$$$i").mkString(", ")
         s"INSERT INTO $tableName ($stringColumns)\nVALUES ($values);"
+      case Delete(tableName, condition) =>
+        val stringCondition = condition.map(generateCondition) match {
+          case Some(conditionString) => List(s"WHERE $conditionString")
+          case None                  => List()
+        }
+        (s"DELETE FROM $tableName" +: stringCondition).mkString("", " ", ";")
     }
 }
--- a/src/main/scala/temple/generate/database/PostgresGenerator.scala
+++ b/src/main/scala/temple/generate/database/PostgresGenerator.scala
@@ -74,22 +74,22 @@ object PostgresGenerator extends DatabaseGenerator {
             .mkString(",\n")
             .pipe(indent(_))
         s"CREATE TABLE $tableName (\n$stringColumns\n);"
-      case Read(tableName, columns, condition) =>
+      case Read(tableName, columns, conditions) =>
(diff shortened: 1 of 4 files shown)


src/test/scala/temple/generate/database/package.scala
```diff
@@ -170,5 +170,22 @@ package object database {
...
         |VALUES ($1, $2, $3, $4, $5, $6, $7);""".stripMargin
+
+    val deleteStatement = Delete(
+      "Users"
+    )
+
+    val postgresDeleteString: String =
+      """DELETE FROM Users;"""
+
+    val deleteStatementWithWhere = Delete(
```
Lewiky, line 181: Are you gonna write a test to do the DELETE with WHERE checking?

Lewiky (changes requested):

src/main/scala/temple/generate/database/PostgresGenerator.scala
```diff
@@ -74,16 +74,22 @@ object PostgresGenerator extends DatabaseGenerator {
...
         s"CREATE TABLE $tableName (\n$stringColumns\n);"
-      case Read(tableName, columns, condition) =>
+      case Read(tableName, columns, conditions) =>
         val stringColumns = columns.map(_.name).mkString(", ")
-        val stringCondition = condition.map(generateCondition) match {
-          case Some(conditionString) => List(s"WHERE $conditionString")
-          case None                  => List()
+        val stringConditions = conditions.map(generateCondition) match {
+          case Some(conditionsString) => List(s"WHERE $conditionsString")
+          case None                   => List()
```
jaylees14, line 81: This is fine, but in the future try and avoid making changes to code that doesn't directly relate to the PR

jaylees14 (approved):
Looks good, just worth adding the tests as @Lewiky suggested.

With respect to your comment about making the type more rigid - this could be possible, but we'd probably end up wrapping lots of separate types in an ADT. Not sure we'd actually gain anything from doing that?

src/test/scala/temple/generate/database/package.scala
c-harding: Will accept when complete

c-harding (changes requested):

c-harding (approved):

src/main/scala/temple/generate/database/PostgresGenerator.scala
willjvsmith: I know what you mean, but this does kinda relate to the PR. Had to write the same stuff for the Delete case and noticed it was wrong.

Lewiky (approved):

PR #17, opened 2020-02-03
[SQL-Gen] Implement DROP TABLE statements

Lewiky:
Allow the generation of Postgres DROP TABLE statements, with or without the IF EXISTS clause. 

Fully tested, also implemented some more tests to increase code-cov.

### DIFF
--- a/src/main/scala/temple/generate/database/PostgresGenerator.scala
+++ b/src/main/scala/temple/generate/database/PostgresGenerator.scala
@@ -91,5 +91,6 @@ object PostgresGenerator extends DatabaseGenerator {
           case None                   => List()
         }
         (s"DELETE FROM $tableName" +: stringConditions).mkString("", " ", ";")
+      case Drop(tableName, ifExists) => s"DROP TABLE $tableName" + { if (ifExists) " IF EXISTS;" else ";" }
     }
 }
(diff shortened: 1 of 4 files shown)


src/main/scala/temple/generate/database/PostgresGenerator.scala
```diff
@@ -91,5 +91,6 @@ object PostgresGenerator extends DatabaseGenerator {
           case None                   => List()
         }
         (s"DELETE FROM $tableName" +: stringConditions).mkString("", " ", ";")
+      case Drop(tableName, ifExists) => s"DROP TABLE $tableName" + { if (ifExists) " IF EXISTS;" else ";" }
```
Lewiky, line 94: Unsure about this - Sticking with the style we've used elsewhere makes tonnes of unnecessary code, but I also don't like the space at the beginning of `" IF EXISTS"`

c-harding: Now this is going to sound weird, but it might work. A function called mkSQL, which takes varargs/a string and some varargs. Each of the varargs can either be a string or a sequence (incl. optional) of strings (using implicit conversion to SQLSubLine?). The function then flattens those varargs to a string, joined with space (and maybe appends a semicolon). Here we can turn the binary into an option (Option.when iirc). There are various places across the generation that could benefit from this.

jaylees14: This is what I initially had in mind for this class - think it'd be quite a big change for this PR so I'd keep it separate. Seems like the right way to go though

jaylees14 (approved):
Looks good, nice job 👍

src/main/scala/temple/generate/database/PostgresGenerator.scala
c-harding: Sounds like fun, would be interested in doing this next sprint

c-harding: Added to backlog

c-harding (approved):
lgt @xsanda

willjvsmith (approved):
LGTM 🚀

PR #19, opened 2020-02-04
[SQL-Gen] Add simple update statement gen

willjvsmith:
* Add simple update statement generation
  * Update all rows in table: `UPDATE Users SET bankBalance = 123.456, name = 'Will';`
  * Update based on condition: `UPDATE Users SET bankBalance = 123.456, name = 'Will' WHERE Users.id = 123456;`
* Add basic unit tests

Expressions are currently just simple string values, but have been written to be extended later as we see need.


c-harding (approved):
All looks good

jaylees14 (approved):
Nice one 🥇

PR #21, opened 2020-02-04
[SBT-Warnings] Turn up and fix compiler warnings

c-harding:
This commit turns up the level of warnings, and then resolves the outstanding warnings

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -22,3 +22,6 @@ lazy val root = (project in file("."))
   )
 
 scalacOptions ++= Seq("-deprecation", "-feature")
+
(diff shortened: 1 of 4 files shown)


c-harding:
Is it possible to add a check that there are no warnings to the CI?

c-harding:
Hope that code coverage is good enough for you guys

jaylees14:
@xsanda We could add a CI build step that uses the `-Xfatal-warnings` flag, but is non blocking. Might be worth doing this as part of the migration to travis?

jaylees14 (approved):
Nice one, thanks @xsanda 👍

Lewiky (approved):
LGTM

PR #22, opened 2020-02-05
[Semantics] Parse the AST into a semantically valid structure

c-harding:

PR #24, opened 2020-02-05
[Test-utils] Test String Utils

c-harding:


c-harding:
Not tested mkString as it has been deleted in #22

jaylees14 (approved):

src/test/scala/temple/utils/StringUtilsTest.scala
```diff
@@ -0,0 +1,22 @@
...
+class StringUtilsTest extends FlatSpec with Matchers {
+  "indent" should "add spaces to an empty string" in {
+    indent("") shouldEqual "  "
+  }
+  "indent" should "add n spaces to an empty string" in {
+    indent("", 1) shouldEqual " "
+    indent("", 3) shouldEqual "   "
+  }
+  "indent" should "add spaces to a single line" in {
+    indent("abcd") shouldEqual "  abcd"
```
willjvsmith, line 15: tests will need to be updated if we ever change the default indent size, but I guess we'll never change it

willjvsmith (approved):
LGTM 🚀

PR #25, opened 2020-02-06
[Floating point parsing] Fix floating point parsing

c-harding:
I miss Yoda, turns out the implicit `try` only wraps the alternation, so
`wholeNumber | floatingPointNumber` will parse `0.0` as `0`, with `.0`
left over.


jaylees14 (approved):
LGTM

Lewiky (approved):
LGTM

PR #26, opened 2020-02-06
[List to Seq] Replace instances of List with superclass Seq

c-harding:
List is a linked list, and converting into this data structure seems a
bit pointless in many cases when they can just be left in the Seq
(immutable ordered collection) that they are already in/can be converted
to most efficiently.

### DIFF
--- a/src/main/scala/temple/DSL/parser/UtilParsers.scala
+++ b/src/main/scala/temple/DSL/parser/UtilParsers.scala
@@ -15,22 +15,22 @@ trait UtilParsers extends RegexParsers {
     * @param q0 The parser marking the end of the sequence. Use `guard` to avoid consuming this input
     * @return A parser that returns a list of the successful parses of `p0`
     */
-  def repUntil[T](p0: => Parser[T], q0: => Parser[Any] = ""): Parser[List[T]] = Parser { in =>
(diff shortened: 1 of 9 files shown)


jaylees14 (approved):
LGTM 💯

src/main/scala/temple/DSL/parser/UtilParsers.scala
```diff
@@ -15,22 +15,22 @@ trait UtilParsers extends RegexParsers {
...
         q(in0) match {
-          case Success(_, next) => Success(elems.toList, next) // allow `q` to consume (use guard to avoid)
-          case _: Failure       => Failure(str, input)         // If both fail, propagate `p0`’s failure.
-          case e: Error         => e                           // still have to propagate `q0`’s parser error
+          case Success(_, next) => Success(elems.toSeq, next) // allow `q` to consume (use guard to avoid)
+          case _: Failure       => Failure(str, input)        // If both fail, propagate `p0`’s failure.
+          case e: Error         => e                          // still have to propagate `q0`’s parser error
         }
-      case _ => Success(elems.toList, in0)
+      case _ => Success(elems.toSeq, in0)
```
c-harding, line 33: This is the source of the poor diff coverage, it’ll be fixed in #22

Lewiky (approved):
LOVE IT  such e f f i c i e n c y

PR #27, opened 2020-02-06
[Protected parser] Make the sub-file parsers protected

c-harding:
This is so that they are only used in the correct scope, as part of a
parser for a Templefile


jaylees14 (approved):

Lewiky (approved):
protected

PR #29, opened 2020-02-08
[infra] Migrate GitHub Actions to Travis

jaylees14:
- Migrate GitHub Actions to Travis
- Break each step into a named travis stage
- Make sure we're actually formatting test & integration test files

NB: This currently runs every stage on every commit - wondering if I should just run the tests on commit then IT/Coverage/Format on PR and merges?


.travis.yml
```diff
@@ -0,0 +1,37 @@
+language: scala
+
+scala:
+    - 2.13.1
```
c-harding: The four spaces do not look like yaml to me

c-harding: And doesn’t this work without being an array?

jaylees14: https://docs.travis-ci.com/user/languages/scala/#specifying-scala-versions

c-harding (approved):
I trust it runs the tests and fails successfully when they don’t pass?

Lewiky (approved):
LGTM

willjvsmith (approved):

PR #30, opened 2020-02-10
[Fix IntelliJ tests] Prevent class naming conflicts

c-harding:
Classes and package objects can’t share names between main, test and it.


c-harding:
Turns out reducing the indent on an entire file breaks git

src/it/scala/temple/containers/DockerPostgresService.scala
```diff
@@ -8,6 +8,23 @@ import temple.containers.DockerPostgresService.PostgresReadyChecker
 import scala.concurrent.{ExecutionContext, Future}
 import scala.concurrent.duration._
 
+/** DockerPostgresService configures a docker container for Postgres */
+trait DockerPostgresService extends DockerKit {
```
c-harding, line 12: Traits need to be declared above their companion object

jaylees14 (approved):
👍

Lewiky (approved):
LGTM

PR #32, opened 2020-02-10
[Collection Utilities] Add utility functions for lists and sequences, along with tests

c-harding:


jaylees14 (approved):
👍 Nice one

src/main/scala/temple/utils/SeqUtils.scala
```diff
@@ -0,0 +1,8 @@
+package temple.utils
+
+object SeqUtils {
+
+  implicit class SeqOptionExtras[A](seq: Seq[Option[A]]) {
+    def sequence: Option[Seq[A]] = if (seq.contains(None)) None else Some(seq.flatten)
```
Lewiky, line 14: Can we get some scaladoc on this? Not obvious what it does

Lewiky (approved):
Small things, could do with a bit more documentation

PR #33, opened 2020-02-11
[Reformat code] Add trailing commas to multiline function calls

c-harding:
Also run IntelliJ’s “reformat code” on everything, which cleans up some imports.


jaylees14 (approved):
LGTM - I'm not too bothered if we have this or not, but if you want it then happy to agree

c-harding:
The main advantage is just that every line inside a multiline function call is the same, so can more easily be reordered, and neighbours do not change when an extra argument is added/removed

willjvsmith (approved):
Also not fussed either way 👍

Lewiky (approved):

PR #34, opened 2020-02-11
[Enumerations] Add enumerations collection

c-harding:
Similar to enumeratum, in that we can enumerate the elements of an enum, but we can also lookup case-insensitively by name, including by aliases


c-harding:
This is yet another part of #22 being broken down

jaylees14 (approved):
Good stuff, shame this isn't supported in Scala by default

willjvsmith (approved):
Nice, LGTM 🚀

PR #37, opened 2020-02-12
[SQL-Gen] Support Blob type in Postgres

jaylees14:
- Add support for Blob in Postgres: [see docs](https://www.postgresql.org/docs/9.1/datatype-binary.html)
- Add unit and integration test to validate generation


src/main/scala/temple/generate/database/PostgresGenerator.scala
```diff
@@ -74,6 +74,7 @@ object PostgresGenerator extends DatabaseGenerator[PostgresContext] {
       case DateCol       => s"DATE"
       case TimeCol       => s"TIME"
       case DateTimeTzCol => s"TIMESTAMPTZ"
+      case BlobCol       => s"BYTEA"
```
Lewiky: 👋☕️

Lewiky (dismissed):
Excellently done

src/it/scala/temple/generate/database/PostgresGeneratorIntegrationTest.scala
```diff
@@ -57,6 +60,30 @@ class PostgresGeneratorIntegrationTest extends PostgresSpec with Matchers with B
     result.isLast shouldBe true
   }
 
+  // Since Blobs require files, we open create a test dynamically
+  it should "correctly store blobs" in {
+    val testImage  = new File("src/it/scala/temple/testfiles/cat.jpeg")
+    val fileStream = new FileInputStream(testImage)
+    try {
+      val fileContents    = fileStream.readAllBytes()
```
c-harding: Can this be moved to a util please?

c-harding (dismissed):
Looks good to go, though a util would be preferable to explicit file IO here

c-harding (approved):
thank you for your service in testing

Lewiky (approved):

PR #38, opened 2020-02-12
[CLI] Add output dir flag to generate option

jaylees14:
Add an output directory flag to place files we generate, defaults to current working directory
Not sure if we should be doing some resolution if they provide a relative paths vs absolute? @xsanda thoughts?


willjvsmith:
Feels a little premature since we don't generate anything to test it with 😢 

I would have thought the IO functions would handle absolute/relative paths for us?

src/main/scala/temple/Main.scala
```diff
@@ -1,10 +1,18 @@
...
 
+import java.nio.file.FileAlreadyExistsException
+
 import temple.DSL.DSLProcessor
 import temple.utils.FileUtils
 
 /** Main entry point into the application */
 object Main extends App {
+
+  def fail(msg: String): Nothing = {
```
Lewiky: Seeing as this is quite similar to a lot of failure stuff Charlie used in #22, would it maybe be worth abstracting this out to a `FailUtils`or something?

Lewiky (approved):
LGTM

src/main/scala/temple/Main.scala
jaylees14: They're a little different - Charlie's throws an exception: https://github.com/TempleEight/temple/blob/3e3f12e87a48b91f85189a6b87c375af47189d5d/src/main/scala/temple/DSL/semantics/ErrorHandling.scala#L19

Might still be worth pulling it out though

src/main/scala/temple/utils/FileUtils.scala
```diff
@@ -1,12 +1,16 @@
 package temple.utils
 
 import java.io.{File, PrintWriter}
+import java.nio.file.{Files, Paths}
```
c-harding, line 4: Can we change io to nio?

src/test/scala/temple/utils/FileUtilsTest.scala
```diff
@@ -0,0 +1,15 @@
+package temple.utils
+
+import java.nio.file.FileAlreadyExistsException
+
+import org.scalatest.{FlatSpec, Matchers}
+
+class FileUtilsTest extends FlatSpec with Matchers {
+  "Creating a folder that already exists" should "not throw an exception" in {
+    FileUtils.createDirectory("src")
```
c-harding, line 14: Can we do this twice, to make it more explicit that the `src` directory exists? Could even wrap the second in `noException should be thrownBy`

c-harding (changes requested):
This has just failed the tests as part of codecov, but not the dedicated test round. This is because the test involves randomly generating a single file path. In the first attempt it went fine, because the string was simple. In the second case, on the other hand, the random string included a `>` character, which is not legal in a file name, so it failed. Firstly the random file name generation needs to be changed, and secondly a single sample of randomness in a test should probably be avoided.

c-harding (approved):

PR #40, opened 2020-02-13
[Travis] Combine test and code cov step

c-harding:
The code cov stage only fails when the tests fail, so they can be combined into a single stage. As the first stage, this means that the codecov bot can report back sooner than when run last.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,8 @@ language: scala
 scala:
   - 2.13.1
 
+os: linux
+
 jdk:
   - oraclejdk11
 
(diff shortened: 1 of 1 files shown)


Lewiky (approved):

c-harding:
@jaylees14 Please could you disable the requirement for Travis’ pull request build, as it’s always identical to the branch’s build?

.travis.yml
```diff
@@ -12,13 +12,11 @@ services:
 jobs:
   include:
     - stage: test
```
jaylees14, line 16: Can you rename the stage to make it clear what is actually being ran?

jaylees14:
> @jaylees14 Please could you disable the requirement for Travis’ pull request build, as it’s always identical to the branch’s build?

Done

jaylees14 (approved):

PR #41, opened 2020-02-13
[Docker-Gen] Basic Dockerfile generation 

Lewiky:
Created the basic infrastructure around Dockerfile generation - currently supports generating just top level statements, but with no parameters. 

The order of the statements is enforced, so that the first line must be a `FROM` clause.

Unit tested, integration tests to come once the full functionality exists.


Lewiky:
Also can update this to be more in adherence with what Charlie did with the `mkSQL` stuff in the Postgres generator

src/test/scala/temple/generate/docker/package.scala
```diff
@@ -0,0 +1,31 @@
...
+        Run,
+        Run,
+        Cmd,
+      ),
+    )
+
+    val basicDockerFileString: String =
+      """|FROM temple/base:1.2.3
+        |
+        |COPY
```
c-harding: Is this a valid sample?

src/test/scala/temple/generate/docker/package.scala
```diff
@@ -0,0 +1,31 @@
+package temple.generate
+
+import temple.generate.docker.ast.DockerfileRoot
+import temple.generate.docker.ast.Statement._
+
+package object docker {
+
+  object TestData {
```
c-harding: is there a reason this is in the package object, not its own file?

src/main/scala/temple/generate/docker/DockerFileGenerator.scala
```diff
@@ -0,0 +1,37 @@
...
+      case Cmd        => "CMD"
+      case Label      => "LABEL"
+      case Expose     => "EXPOSE"
+      case Add        => "ADD"
+      case Copy       => "COPY"
+      case EntryPoint => "ENTRYPOINT"
+      case Volume     => "VOLUME"
+      case User       => "USER"
+      case WorkDir    => "WORKDIR"
+      case Arg        => "ARG"
```
jaylees14: Surely these have arguments?

jaylees14 (commented):
Not too sure I'm following this PR. If you're trying not to implement too much, remove the case classes and add them later? So fully support a subset of all possible commands for now?

src/test/scala/temple/generate/docker/DockerFileGeneratorTest.scala
```diff
@@ -0,0 +1,13 @@
+package temple.generate.docker
+
+import org.scalatest.{FlatSpec, Matchers}
+
+class DockerFileGeneratorTest extends FlatSpec with Matchers {
+
+  behavior of "DockerFileGeneratorTest"
```
c-harding: This is doing nothing, either use
```scala
behaviour of "Thing"
it should ???
```
or
```scala
"Thing" should ???
```
but not both

src/main/scala/temple/generate/docker/DockerFileGenerator.scala
```diff
@@ -0,0 +1,37 @@
+package temple.generate.docker
+
+import temple.generate.docker.ast.Statement._
+import temple.generate.docker.ast.{DockerfileRoot, Statement}
+
+/** Generator object for building Dockerfiles from the Dockerfile AST objects */
+object DockerFileGenerator {
+
+  /** Given a [ast.Statement], generate a valid string */
```
c-harding: `Given a [[temple.generate.database.ast.Statement]],`

src/main/scala/temple/generate/docker/DockerFileGenerator.scala
```diff
@@ -0,0 +1,37 @@
...
+  private def generateStatement(statement: Statement): String =
+    statement match {
+      case Env => "ENV"
+      case From(image, tag) =>
+        s"FROM $image" + (tag match {
+          case Some(tagString) => s":$tagString"
+          case None            => ""
+        })
```
c-harding: This would be equivalent to one of the following: I feel `mkSQL` should be renamed and moved if this were to happen
```scala
mkSQL("FROM", mkSQL.noSpaces(image, tag.map(":" + _))) // With a noSpaces method added
mkSQL("FROM", image, tag.map(":" + _)) // With colon set as a closing punctuation mark
```

src/main/scala/temple/utils/StringUtils.scala
```diff
@@ -5,4 +5,10 @@ object StringUtils {
 
   /** Given a string, indent it by a given number of spaces */
   def indent(str: String, length: Int = 2): String = str.replaceAll("^|(?<=\n)", " " * length)
+
+  /** Given a string and a wrapping char, append the char at each end */
+  def wrap(str: String, char: Char = '"'): String = s"$char$str$char"
+
+  /** Given a string that may contain quotes, automatically escape them */
+  def escapeQuotes(str: String): String = str.split('"').mkString("\\\"")
```
c-harding: Firstly, this doesn't escape backslashes. Secondly, surely this has been done before?

Lewiky: How do you mean done before? In a library?

Lewiky: And how much santisation do we need? Surely if this is an issue here it's an issue in all the other generation code?

src/main/scala/temple/generate/docker/DockerFileGenerator.scala
Lewiky: This is a good point - is there anything in `SQLTerm` that is actually specific to SQL generation? Could be moved to a general `generate.utils` package that does all sorts of helpful stuff for generation like this

src/main/scala/temple/utils/StringUtils.scala
c-harding: I mean that there will be a library function for escaping this stuff. And it's more important here than elsewhere, because everything else I've done has been from identifiers parsed from the source, not free-form command-line input, where it's realistic that unusual characters could appear.

However, this here is simply a JSON array, so use an off-the-shelf JSON generator instead of `buildArrayString`

Lewiky: Sure I'll do that - Though I don't think the input here would be command-line input - it'd be generated from the parsed source same as the postgres and go gen

src/main/scala/temple/generate/docker/DockerFileGenerator.scala
c-harding: I don’t believe so. It should apply to any whitespace-independent language, though it is opinionated about when to insert spaces (not after `(` or whitespace, and not before `,;)` or whitespace). You'd need to add colon to that, and probably `[{`/`}]` too, as well as possibly adding a noSpaces method

src/main/scala/temple/utils/StringUtils.scala
c-harding: Not from command-line input, but it contains command-line commands, which frequently contain “special” characters (in the RUN block)

src/main/scala/temple/utils/StringUtils.scala
```diff
@@ -5,4 +5,10 @@ object StringUtils {
 
   /** Given a string, indent it by a given number of spaces */
   def indent(str: String, length: Int = 2): String = str.replaceAll("^|(?<=\n)", " " * length)
+
+  /** Given a string and a wrapping char, append the char at each end */
+  def wrap(str: String, char: Char = '"'): String = s"$char$str$char"
```
c-harding: This should just be JSON-encoding specialised to strings, right?

src/main/scala/temple/generate/docker/DockerFileGenerator.scala
Lewiky: Also you think it's worth adding a whole JSON lib dependency for this? Scala doesn't have JSON processing built in

src/main/scala/temple/utils/StringUtils.scala
Lewiky: Yeah that's true 

Do you think it's worth adding a dependency on a  whole JSON library for this?

c-harding: I do, as it will cover edge cases, as we’ll almost certainly need to generate JSON in the future

src/test/scala/temple/generate/docker/UnitTestData.scala
```diff
@@ -0,0 +1,23 @@
...
+    ),
+  )
+
+  val basicDockerFileString: String =
+    """|FROM temple/base:1.2.3
+        |
+        |RUN /bin/bash
+        |
+        |RUN ["/bin/python3", "src/main.py", "--help"]
+        |""".stripMargin
```
jaylees14: Think the whitespace is slightly off here

src/main/scala/temple/generate/docker/DockerFileGenerator.scala
```diff
@@ -0,0 +1,30 @@
+package temple.generate.docker
+
+import io.circe.Printer
+import io.circe.syntax._
+import temple.generate.docker.ast.Statement._
+import temple.generate.docker.ast.{DockerfileRoot, Statement}
+import temple.generate.utils.CodeTerm.mkCode
+
+/** Generator object for building Dockerfiles from the Dockerfile AST objects */
+object DockerFileGenerator {
```
jaylees14: Just a consistency one, think we should go with `Dockerfile` rather than `DockerFile`

jaylees14 (approved):

c-harding (approved):
looks good to go

PR #42, opened 2020-02-13
[Code-Formatting] Refactor `SQLTerm` to be generic

Lewiky:
While working on #41, was pointed out that `SQLTerm` File isn't specific to just SQL generation and is generally useful elsewhere. 

This PR simply moves it into a `generate.utils` package and renames all of the variables away from SQL based names, to generic `Code` based ones.


src/main/scala/temple/generate/utils/CodeTerm.scala
```diff
@@ -0,0 +1,98 @@
...
+
+  /** Turns a [[String]] into a [[CodeTerm]], with a trivial iterator of just the single string
+    */
+  implicit class CodeTermString(string: String) extends CodeTerm {
+    override def flatIterator: Iterator[String] = Iterable.single(string).iterator
+  }
+
+  /**
+    * Turns a list of items into a string with separators, constructed with
+    * [[CodeTerm.mkCode#list(temple.generate.database.CodeTerm.CodeTermIterableString)]]
```
c-harding: Is this correct?

c-harding: Right-click on the function, copy reference

c-harding (approved):
Looks fine, though GitHub’s similarity threshold is too high to properly see what's changed

jaylees14 (approved):

PR #43, opened 2020-02-13
Test naming

jaylees14:
To improve the output of `sbt test`, tests for the same functionality can be grouped together using `behavior of ...`.

This works well when the tests in the same file test the same behaviour, but not when there's multiple behaviours being tested (see the integration tests) - are we happy to stick with this, or should I rename the tests to contain the behaviour in the right side of the `should`, so that the left side can always remain the same?

### DIFF
--- a/src/test/scala/temple/generate/database/PostgresGeneratorTest.scala
+++ b/src/test/scala/temple/generate/database/PostgresGeneratorTest.scala
@@ -12,49 +12,64 @@ class PostgresGeneratorTest extends FlatSpec with Matchers {
   it should "generate correct CREATE statements" in {
     PostgresGenerator.generate(TestData.createStatement) shouldBe TestData.postgresCreateString
   }
+
   it should "generate correct SELECT statements" in {
     PostgresGenerator.generate(TestData.readStatement) shouldBe TestData.postgresSelectString
   }
+
   it should "handle column constraints correctly" in {
     PostgresGenerator.generate(TestData.createStatementWithConstraints) shouldBe TestData.postgresCreateStringWithConstraints
   }
+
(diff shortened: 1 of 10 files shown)


Lewiky (approved):
I personally like it - see what you mean about the Integration tests being a bit tricky cause they aren't testing the behaviour of one particular component..  I think we should either mix the styles between test types (boo) or stick with this, it's not the end of the world

src/test/scala/temple/generate/database/PostgresGeneratorTest.scala
```diff
@@ -7,53 +7,55 @@ class PostgresGeneratorTest extends FlatSpec with Matchers {
 
   implicit val context: PostgresContext = PostgresContext(PreparedType.DollarNumbers)
 
-  "PostgresGenerator" should "generate correct CREATE statements" in {
+  behavior of "PostgresGenerator"
```
c-harding, line 10: Can this file have some blank lines between tests please?

src/test/scala/temple/utils/StringUtilsTest.scala
```diff
@@ -4,18 +4,19 @@ import org.scalatest.{FlatSpec, Matchers}
 import temple.utils.StringUtils.indent
 
 class StringUtilsTest extends FlatSpec with Matchers {
-  "indent" should "add spaces to an empty string" in {
+  behavior of "indent"
+  it should "add spaces to an empty string" in {
```
c-harding, line 9: Can this file have some blank lines between tests please?

c-harding (approved):
Would be nice to have consistent spacing, but looks good

PR #46, opened 2020-02-14
[Semantics] Make all attribute type parameters optional

c-harding:
At @jaylees14’s request, string `minLength` and float `min` and `max` now have None as a default value, rather than `0`/`MinValue`/`MaxValue` respectively


jaylees14 (approved):
❤️

Lewiky (approved):

PR #49, opened 2020-02-14
[SQL-Gen] Add support for VARCHAR types

jaylees14:
Add support for VARCHAR strings, when an upper bound is enforced


Lewiky (approved):

c-harding (approved):
Very simple, if repetitive. Nice

PR #50, opened 2020-02-14
[Docker-Gen]  Implement more Dockerfile Commands

Lewiky:
Implement a load more Docker commands

- ENV
- ADD
- COPY
- ENTRYPOINT
- VOLUME
- WORKDIR

Unit tested


src/main/scala/temple/generate/docker/DockerfileGenerator.scala
```diff
@@ -16,10 +16,16 @@ object DockerfileGenerator {
...
-      case Run(executable, params) => mkCode("RUN", buildArrayString(executable +: params))
-      case Cmd(executable, params) => mkCode("CMD", buildArrayString(executable +: params))
-      case Expose(port)            => mkCode("EXPOSE", port.toString)
+      case From(image, tag)               => mkCode("FROM", image, tag.map(":" + _))
+      case Run(executable, params)        => mkCode("RUN", buildArrayString(executable +: params))
+      case Cmd(executable, params)        => mkCode("CMD", buildArrayString(executable +: params))
+      case Expose(port)                   => mkCode("EXPOSE", port.toString)
+      case Env(key, value)                => mkCode("ENV", key, value)
+      case Add(src, dest)                 => mkCode("ADD", src, dest)
+      case Copy(src, dest)                => mkCode("COPY", src, dest)
```
c-harding, line 25: Does this work with a space in the file path? Could that ever happen?

c-harding (commented):
Looks good, just concerned about parsing errors

src/main/scala/temple/generate/docker/DockerfileGenerator.scala
Lewiky: Well to be a valid file path in the first place, any spaces would need to be escaped, which would get propagated through to the Dockerfile, which also supports escaped spaces in paths.

c-harding: Not sure that’s true, but if you’re sure they’ve already been fully escaped then that’s fine

c-harding (approved):

jaylees14 (approved):

PR #51, opened 2020-02-16
[SQL-Gen] Map from semantic types into database creation types

jaylees14:
This is a first in a number of PRs that perform the mapping from our Templefile to Postgres. Here, we perform the highest level mapping, from `ServiceBlock` to `Statement.Create`.

The implementation here is database agnostic - the next PR will call these functions and pass the result to the associated database provider. 

I’m not 100% set on the `Builder` name, please shout if you have a better suggestion.

The current package structure I’m envisaging looks like this - where the main mapping entry point is `ProjectBuilder`, which will assemble an entire project by calling into the generic sub-builders, and then calling into the implementation specific objects, like `PostgresGenerator` etc. Thoughts?:

```
📁 temple
┖─ 📁 builder
   ┠─ 📝 DatabaseBuilder.scala
   ┠─ 📝 DockerBuilder.scala
   ┠─ 📝 ServerBuilder.scala
   ┠─ ...
   ┖─ 📁 project
       ┠─ 📝 ProjectBuilder.scala
       ┖─ ...
```

### Test Plan
- Additional unit tests added, with simple types and complex constrained types

### DIFF
--- a/src/main/scala/temple/builder/DatabaseBuilder.scala
+++ b/src/main/scala/temple/builder/DatabaseBuilder.scala
@@ -18,7 +18,7 @@ object DatabaseBuilder {
       case Annotation.Unique => ColumnConstraint.Unique
     }.toSeq
 
-    val (colType, constraints) = attribute.attributeType match {
+    val (colType, typeConstraints) = attribute.attributeType match {
       case AttributeType.BoolType     => (ColType.BoolCol, Nil)
       case AttributeType.DateType     => (ColType.DateCol, Nil)
       case AttributeType.DateTimeType => (ColType.DateTimeTzCol, Nil)
@@ -34,7 +34,7 @@ object DatabaseBuilder {
         val colType = if (max.isDefined) ColType.BoundedStringCol(max.get) else ColType.StringCol
         (colType, generateMaxMinConstraints(s"length($name)", None, min))
     }
-    ColumnDef(name, colType, valueConstraints ++ constraints)
+    ColumnDef(name, colType, valueConstraints ++ typeConstraints)
   }
 
   /**
--- a/src/main/scala/temple/builder/DatabaseBuilder.scala
+++ b/src/main/scala/temple/builder/DatabaseBuilder.scala
@@ -19,21 +19,22 @@ object DatabaseBuilder {
     }.toSeq
 
     val (colType, constraints) = attribute.attributeType match {
-      case AttributeType.BoolType     => (ColType.BoolCol, valueConstraints)
-      case AttributeType.DateType     => (ColType.DateCol, valueConstraints)
-      case AttributeType.DateTimeType => (ColType.DateTimeTzCol, valueConstraints)
-      case AttributeType.TimeType     => (ColType.TimeCol, valueConstraints)
-      case AttributeType.ForeignKey   => (ColType.IntCol(4), valueConstraints)
-      case AttributeType.BlobType(_)  => (ColType.BlobCol, valueConstraints)
+      case AttributeType.BoolType     => (ColType.BoolCol, Nil)
+      case AttributeType.DateType     => (ColType.DateCol, Nil)
+      case AttributeType.DateTimeType => (ColType.DateTimeTzCol, Nil)
+      case AttributeType.TimeType     => (ColType.TimeCol, Nil)
+      case AttributeType.ForeignKey   => (ColType.IntCol(4), Nil)
+      case AttributeType.BlobType(max) =>
+        (ColType.BlobCol, generateMaxMinConstraints(s"octet_length($name)", max, None))
       case AttributeType.IntType(max, min, precision) =>
-        (ColType.IntCol(precision), valueConstraints ++ generateMaxMinConstraints(name, max, min))
+        (ColType.IntCol(precision), generateMaxMinConstraints(name, max, min))
       case AttributeType.FloatType(max, min, precision) =>
-        (ColType.FloatCol(precision), valueConstraints ++ generateMaxMinConstraints(name, max, min))
+        (ColType.FloatCol(precision), generateMaxMinConstraints(name, max, min))
       case AttributeType.StringType(max, min) =>
         val colType = if (max.isDefined) ColType.BoundedStringCol(max.get) else ColType.StringCol
-        (colType, valueConstraints ++ generateMaxMinConstraints(s"length($name)", None, min))
+        (colType, generateMaxMinConstraints(s"length($name)", None, min))
     }
-    ColumnDef(name, colType, constraints)
+    ColumnDef(name, colType, valueConstraints ++ constraints)
   }
 
   /**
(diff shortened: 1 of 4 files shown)


src/main/scala/temple/builder/DatabaseBuilder.scala
```diff
@@ -0,0 +1,57 @@
...
+      case Annotation.Unique => ColumnConstraint.Unique
+    }.toSeq
+
+    val (colType, constraints) = attribute.attributeType match {
+      case AttributeType.BoolType     => (ColType.BoolCol, valueConstraints)
+      case AttributeType.DateType     => (ColType.DateCol, valueConstraints)
+      case AttributeType.DateTimeType => (ColType.DateTimeTzCol, valueConstraints)
+      case AttributeType.TimeType     => (ColType.TimeCol, valueConstraints)
+      case AttributeType.ForeignKey   => (ColType.IntCol(4), valueConstraints)
+      case AttributeType.BlobType(_)  => (ColType.BlobCol, valueConstraints)
```
c-harding: Is there a reason the blob size is not enforced?

src/main/scala/temple/builder/DatabaseBuilder.scala
```diff
@@ -0,0 +1,57 @@
+package temple.builder
+
+import temple.DSL.semantics.{Annotation, Attribute, AttributeType, ServiceBlock}
+import temple.generate.database.ast.ColumnConstraint.Check
+import temple.generate.database.ast._
+
+/** Construct database queries from a Templefile structure */
+object DatabaseBuilder {
+
+  private def generateMaxMinConstraints[T](name: String, max: Option[T], min: Option[T]): Seq[ColumnConstraint] = {
```
c-harding, line 10: `[T: Numeric]` may be useful as the type parameter

src/main/scala/temple/builder/DatabaseBuilder.scala
```diff
@@ -0,0 +1,57 @@
...
+      case AttributeType.DateTimeType => (ColType.DateTimeTzCol, valueConstraints)
+      case AttributeType.TimeType     => (ColType.TimeCol, valueConstraints)
+      case AttributeType.ForeignKey   => (ColType.IntCol(4), valueConstraints)
+      case AttributeType.BlobType(_)  => (ColType.BlobCol, valueConstraints)
+      case AttributeType.IntType(max, min, precision) =>
+        (ColType.IntCol(precision), valueConstraints ++ generateMaxMinConstraints(name, max, min))
+      case AttributeType.FloatType(max, min, precision) =>
+        (ColType.FloatCol(precision), valueConstraints ++ generateMaxMinConstraints(name, max, min))
+      case AttributeType.StringType(max, min) =>
+        val colType = if (max.isDefined) ColType.BoundedStringCol(max.get) else ColType.StringCol
```
c-harding, line 34: This feels safer than using `.get`:
`max.fold(ColType.StringCol) { ColType.BoundedStringCol(_) }`

src/main/scala/temple/builder/DatabaseBuilder.scala
```diff
@@ -0,0 +1,57 @@
...
+    Seq(maxCondition, minCondition).flatten
+  }
+
+  private def toColDef(name: String, attribute: Attribute): ColumnDef = {
+    val valueConstraints = attribute.valueAnnotations.map {
+      case Annotation.Unique => ColumnConstraint.Unique
+    }.toSeq
+
+    val (colType, constraints) = attribute.attributeType match {
+      case AttributeType.BoolType     => (ColType.BoolCol, valueConstraints)
```
c-harding: Personally I feel that using `Nil` here as the `typeConstraint`, and adding on `valueConstraints` afterwards, would be less repetitive than including `valueConstraints` on every line

src/main/scala/temple/builder/DatabaseBuilder.scala
```diff
@@ -0,0 +1,57 @@
...
+
+  /**
+    * Converts a service block to an associated list of database table create statement
+    * @param serviceName The service name
+    * @param service The ServiceBlock to generate
+    * @return the associated create statement
+    */
+  def createServiceTables(serviceName: String, service: ServiceBlock): Seq[Statement.Create] = {
+    // Create a list of all tables to be created, using top level attributes and nested structs
+    val allTables = service.structs.map(s => (s._1, s._2.attributes)).toSeq :+ (serviceName, service.attributes)
```
c-harding, line 48: I wonder, do you think a `flatten(serviceName: String): Map[String, Map[String, Attribute]]` function would be useful on the ServiceBlock?

src/main/scala/temple/builder/DatabaseBuilder.scala
```diff
@@ -0,0 +1,57 @@
...
+
+    // Generate the create statement for each table
+    allTables
+      .map {
```
c-harding: Why’s this on two lines?

src/main/scala/temple/builder/DatabaseBuilder.scala
```diff
@@ -0,0 +1,57 @@
...
+        case (tableName, attributes) =>
+          val columns = attributes.map(a => toColDef(a._1, a._2))
```
c-harding:
```scala
val columns = attributes.map { case (name, attribute) => toColDef(name, attribute) }
```

c-harding (dismissed):
A couple of suggestions, but looks good

src/main/scala/temple/builder/DatabaseBuilder.scala
jaylees14: I know the Postgres implementation doesn’t use a length, so it was never added

src/main/scala/temple/builder/DatabaseBuilder.scala
jaylees14: Don’t think that’s exactly true - we wouldn’t gain anything from it, and in the future we wouldn’t be able to produce constraints over two variables

```
CHECK val_a >= val_b
```

src/main/scala/temple/builder/DatabaseBuilder.scala
c-harding: https://github.com/jOOQ/jOOQ/issues/6289#issuecomment-304618975

You can `CHECK` the `octet_length`

c-harding: Also, [jOOQ](https://github.com/jOOQ/jOOQ) looks useful

c-harding: Though they haven’t got around to implementing that yet

jaylees14: Nice, shall add that

src/main/scala/temple/builder/DatabaseBuilder.scala
jaylees14: I'm getting a compiler error if I try this, even when I annotate the type of `colType`

jaylees14: As an aside, the implementation of `fold` is exactly what I wrote originally - checks if it's empty and uses `.get` otherwise

c-harding: Never mind then

src/main/scala/temple/builder/DatabaseBuilder.scala
```diff
@@ -19,21 +19,22 @@ object DatabaseBuilder {
     }.toSeq
 
     val (colType, constraints) = attribute.attributeType match {
```
c-harding: Can we rename `constraints` to something more specific, to justify it not including `valueConstraints`?

willjvsmith (approved):
LGTM 🚀

c-harding (approved):
lgtm

Lewiky:
Feedback on directory structure - don't think having the entry point `ProjectBuilder` being a subpackage of the other builders works semantically, the others should be in a subpackage of the root

PR #54, opened 2020-02-17
[builder] Project Builder

jaylees14:
Now map one level higher... `Templefile` to `Project`, which will eventually contain all of the code we should output.

I'll do a follow up PR after this to:
1. Address issues raised offline about file structure
2. Do the final integration with file IO in `Main`

### DIFF
--- a/src/main/scala/temple/builder/project/ProjectBuilder.scala
+++ b/src/main/scala/temple/builder/project/ProjectBuilder.scala
@@ -10,6 +10,11 @@ import temple.generate.database.ast.Statement
 
 object ProjectBuilder {
 
+  /**
+    * Converts a Templefile to an associated project, containing all generated code
+    * @param templefile The semantically correct Templefile
+    * @return the associated generated project
+    */
   def build(templefile: Templefile): Project = {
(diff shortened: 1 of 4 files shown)


src/main/scala/temple/builder/project/ProjectBuilder.scala
```diff
@@ -0,0 +1,37 @@
...
+  /**
+    * Converts a Templefile to an associated project, containing all generated code
+    * @param templefile The semantically correct Templefile
+    * @return the associated generated project
+    */
+  def build(templefile: Templefile): Project = {
+    val databaseCreationScripts = templefile.services.map {
+      case (name, service) =>
+        val queries: Seq[Statement.Create] = DatabaseBuilder.createServiceTables(name, service)
+        val generatedScripts: Seq[String] = service.lookupMetadata[Database] match {
```
c-harding:
```scala
service.lookupMetadata[Database].getOrElse(Postgres)
```

Lewiky (dismissed):
lgtm

c-harding (approved):
`<3`

Lewiky (approved):

PR #62, opened 2020-02-17
[AST] Improve the test failure messages for parse failures

c-harding:
Before:
```
false was not equal to true
ScalaTestFailureLocation: temple.DSL.parser.DSLParserTest at (DSLParserTest.scala:24)
Expected :true
Actual   :false
<Click to see difference>
```

After:
```
Parse error: ':' expected but '}' found
1 | Test: service {x }
                     ^
ScalaTestFailureLocation: temple.DSL.parser.DSLParserTest$ParseResult at (DSLParserTest.scala:13)
```


Lewiky (approved):
This is hot

jaylees14 (approved):

PR #67, opened 2020-02-18
[Infra] Skip assembling in preparation for deployment

c-harding:
Previously the only conditional part was on the deployment part of the deploy stage, rather than the whole deploy stage. This makes Travis faster 😃


jaylees14 (approved):

willjvsmith (approved):
LGTM 🚀

PR #70, opened 2020-02-18
[DSL] Support the Enumerate metadata item

c-harding:
* Add support for parsing the metadata `#enumerate(by: User)`
* Add support for optional arguments for metadata (e.g. `by` in `#enumerate`)
* Add support for omitting the parentheses for metadata without arguments (`#enumerate;` as a shorthand for `#enumerate();`)

As asides:

* Fix the ordering of the semicolon parser, so it says `expected ";"` rather than `expected regex "$"` when something other than the end of the line is reached
* Adds methods for generating square and curly brackets

### DIFF
--- a/src/main/scala/temple/DSL/semantics/MetadataParser.scala
+++ b/src/main/scala/temple/DSL/semantics/MetadataParser.scala
@@ -50,9 +50,7 @@ class MetadataParser[T <: Metadata]() {
     * Add a handler for a new type of metadata, with an optional argument.
     *
     * @param metaKey The name of the metadata item to add
-    * @param argKey The name of the single argument to the metadata. Note that there is also
-    *               [[temple.DSL.semantics.MetadataParser#registerOptionalKeyword(java.lang.String, java.lang.String, temple.DSL.semantics.ArgType, scala.Function1)]]
-    *               if this is the same as `metaKey`.
+    * @param argKey The name of the single argument to the metadata.
     * @param argType The type of the field to expect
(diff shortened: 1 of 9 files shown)


src/main/scala/temple/DSL/semantics/MetadataParser.scala
```diff
@@ -39,13 +40,39 @@ class MetadataParser[T <: Metadata]() {
...
+        implicit val context: Context = Context(metaKey)
+        val argMap                    = parseParameters(argKey -> Some(NoArg))(args)
+        constructor(argMap.getOptionArg(argKey, argType))
+      })
+
+  /** A shorthand for
+    * [[temple.DSL.semantics.MetadataParser#registerOptionalKeyword(java.lang.String, temple.DSL.semantics.ArgType, scala.Function1)]]
+    * with the same `key` used for both the metadata name and its single argument */
+  final protected def registerOptionalKeyword[A](key: String, argType: ArgType[A])(constructor: Option[A] => T): Unit =
+    registerOptionalKeyword(key, key, argType)(constructor)
```
c-harding: This line has no coverage, I included it purely for consistency. Should I remove it?

jaylees14: Yeah I think so, it's easy to add in the future if we need it

jaylees14 (approved):
👍 LGTM

Lewiky (approved):
LGTM

PR #71, opened 2020-02-19
[DSL] Add support for comments in the DSL

c-harding:
Sorry it’s in Regex, it’s just how [scala-parser-combinators](https://www.javadoc.io/static/org.scala-lang.modules/scala-parser-combinators_2.12/1.1.2/scala/util/parsing/json/Lexer.html#whitespace:Lexer.this.Parser[List[Lexer.this.Elem]]) is implemented. I hope the comment explains it adequately: anywhere where whitespace can occur, so can a comment. Note that this does not include places such as inside a string or a comment, as whitespace characters there are treated as character literals rather than ignored as whitespace.

### DIFF
--- a/src/main/scala/temple/DSL/parser/DSLParser.scala
+++ b/src/main/scala/temple/DSL/parser/DSLParser.scala
@@ -2,6 +2,7 @@ package temple.DSL.parser
 
 import temple.DSL.syntax._
 
+import scala.util.matching.Regex
 import scala.util.parsing.combinator.JavaTokenParsers
 
 /** A library of parser generators for the Templefile DSL */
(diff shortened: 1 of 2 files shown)


jaylees14 (approved):
Hmm, not a huge fan of regular expressions, but it does the job and you've added the comment.

Lewiky (approved):
Awesome

c-harding:
> Hmm, not a huge fan of regular expressions, but it does the job and you've added the comment.

@jaylees14 while I can’t say I’m not, the `whitespace` parser is part of `scala-parser-combinators`: it’s inserted between every pair of symbols. As such, the only way around it would be to make a different parser and manually insert it between every pair of combinators, which will get disgusting

PR #72, opened 2020-02-19
Dockerfile integration tests

jaylees14:
This PR adds support for validating and linting Dockerfiles. 
There are two options available to do this:
- Run Docker as a Docker container, feed it the Dockerfile and see what it does
- Use an external linter and trust it has the correct validation

Since the first one would be hard to determine what "correct" behaviour is, I have opted for the second, in the form of [Hadolint](https://github.com/hadolint/hadolint), a Dockerfile linter.

It turns out the existing framework for integration testing only supports interaction with the container via HTTP or Web Sockets [1], so, in order to make requests to Hadolint, it had to be invoked as a result of a web request. This was done via [shell2http](https://github.com/msoap/shell2http), which is wrapped inside of a Docker container, alongside Hadolint.

When a request is made to `http://localhost:8080/hadolint?dockerfile="{"contents":"FROM ...."}"`, the binary is invoked and any errors are returned. These errors follow the rules outlined [here](https://github.com/hadolint/hadolint#rules). The contents of the `Dockerfile` parameter is a JSON object where the key is `contents` and value is the raw Dockerfile. This is decoded in the container using [jq](https://stedolan.github.io/jq/). 

In addition to the framework only supporting HTTP requests, it only supports Docker images that can be pulled from an external repository and not those built locally. Therefore, the Dockerfile containing `shell2http` and `hadolint` was uploaded to my personal Docker Hub account. I've included the deployment script in this PR.

Finally, I've added some sample tests to show how the validation script works. I'll do a follow up PR now that this infrastructure is in place that contains actual test cases of our code.

[1] https://github.com/whisklabs/docker-it-scala#setup 

#### Test Plan

```
sbt:temple> it:test
...
[info] All tests passed.
```


docker/hadolint.Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM hadolint/hadolint:latest AS hadolint
```
c-harding: latest?

src/it/scala/temple/containers/DockerHadolintService.scala
```diff
@@ -0,0 +1,38 @@
...
+import scala.concurrent.{ExecutionContext, Future}
+import scala.concurrent.duration._
+
+/** DockerHadolintService configures a docker container for hadolint */
+trait DockerHadolintService extends DockerKit {
+
+  val hadolintContainer: DockerContainer = DockerContainer(DockerHadolintService.image)
+    .withPorts(DockerHadolintService.internalPort -> Some(DockerHadolintService.externalPort))
+    .withReadyChecker(
+      new HadolintReadyChecker().looped(5, 10.seconds),
```
c-harding, line 16: I maintain this is BS

docker/hadolint.Dockerfile
c-harding: latest?

src/it/scala/temple/containers/HadolintSpec.scala
```diff
@@ -0,0 +1,25 @@
...
+import io.circe.syntax._
+import org.scalatest.{BeforeAndAfterAll, FlatSpec}
+import scalaj.http.Http
+
+trait HadolintSpec extends FlatSpec with DockerTestKit with DockerHadolintService with BeforeAndAfterAll {
+  implicit override val dockerFactory: DockerFactory = new SpotifyDockerFactory(DefaultDockerClient.fromEnv().build())
+
+  // https://github.com/hadolint/hadolint#rules
+  object Rules {
+    val useLatest = "DL3007"
```
c-harding: Isn’t something complaining about the lack of type annotation?

src/it/scala/temple/generate/docker/DockerGeneratorIntegrationTest.scala
```diff
@@ -0,0 +1,16 @@
...
+
+class DockerGeneratorIntegrationTest extends HadolintSpec with Matchers with BeforeAndAfter {
+
+  behavior of "DockerValidator"
+  it should "fail when a latest tag is used" in {
+    validate("FROM golang:latest\nEXPOSE 80") should include(Rules.useLatest)
+  }
+
+  it should "succeed when a versioned tag is used" in {
+    validate("FROM golang:123") shouldBe empty
```
c-harding: Can this expose 80 for consistency (to avoid adding another source of bugs? It also feels like this should be before the other one

src/it/scala/temple/containers/DockerHadolintService.scala
jaylees14: 🤷‍♂

Lewiky (approved):

c-harding (approved):

PR #79, opened 2020-02-20
Allow metadata in structs

### LINKED ISSUE #78: Metadata in struct blocks not parsing correctly
When running `run generate -o my-project src/test/scala/temple/testfiles/simple.temple`, an error is thrown:

```
Found metadata in struct block (Fred): `#enumerable (by: friend);`
```

We should be unifying the test files, and fixing this issue

c-harding:
- Allow metadata in structs (currently only #enumerable) (Fixes #78)
- Add an end to end test for parsing the simple Templefile
- Use the same `simple.temple` file for both sets of tests


jaylees14 (approved):

Lewiky (approved):
LGTM

PR #81, opened 2020-02-20
[Semantics] Boost code coverage of semantic parser

c-harding:
The test to cover these lines contained two errors: it was supposed to throw an error that there were too many arguments, but instead it was throwing an error that the wrong field was given, and there weren’t too many arguments anyway. Fixing this should get code coverage up to the sky.

https://github.com/TempleEight/temple/blob/def8b87d4c57ed40fb37d69867d4f6e98f12ea84/src/main/scala/temple/DSL/semantics/Analyser.scala#L30-L31

Side effect: it turns out IntelliJ doesn’t run tests when there is anything else in the same class (e.g. functions/imports) so I’ve fixed that. That’s why I missed that the first commit wasn’t passing the tests locally.


Lewiky (approved):

src/test/scala/temple/DSL/semantics/SemanticAnalyserTest.scala
```diff
@@ -6,23 +6,11 @@ import temple.DSL.semantics.AttributeType._
 import temple.DSL.syntax
 import temple.DSL.syntax.Arg._
 import temple.DSL.syntax.{Args, DSLRootItem, Entry}
+import temple.DSL.syntax.Entry.{Attribute => AttributeEntry}
```
c-harding: Renamed to avoid this line wrapping onto
```scala
Entry
  .Attribute(
    "a",
    // ...
```
https://github.com/TempleEight/temple/pull/81/files#diff-1a0467d6378d04f12e4aa1a1be41543fR110

jaylees14: This feels like a lot of unnecessary faff for a bit of line wrapping

jaylees14 (approved):
IntelliJ is stupid - long live Vim

PR #82, opened 2020-02-21
[test] Add tests for parsing keyword arguments in any order

c-harding:
Fixes (from Lewis’ mind import bug), checks that `int(min: 5, max: 6)` works just as well as `int(max: 6, min: 5)`


Lewiky (approved):
YE ROY

jaylees14 (approved):
Noice

PR #83, opened 2020-02-21
Table naming

jaylees14:
Renamed all services to be singular (i.e `Users` → `User`) - however, since User is a reserved keyword in Postgres and we don't have something to handle that, I've renamed it to `TempleUser`.

All table names are now in snake case, so we have a method to convert user provided names to snake case

### DIFF
--- a/src/main/scala/temple/utils/StringUtils.scala
+++ b/src/main/scala/temple/utils/StringUtils.scala
@@ -7,6 +7,7 @@ object StringUtils {
   def indent(str: String, length: Int = 2): String = str.replaceAll("^|(?<=\n)", " " * length)
 
   /** Given a string, convert it to snake case */
+  // https://github.com/lift/framework/blob/f1b450db2dd6a22cf9ffe5576ec34c8e87118319/core/util/src/main/scala/net/liftweb/util/StringHelpers.scala#L91
   def snakeCase(str: String): String =
     str
       .replaceAll("([A-Z]+)([A-Z][a-z])", "$1_$2")
--- a/src/main/scala/temple/utils/StringUtils.scala
+++ b/src/main/scala/temple/utils/StringUtils.scala
@@ -5,4 +5,11 @@ object StringUtils {
 
   /** Given a string, indent it by a given number of spaces */
   def indent(str: String, length: Int = 2): String = str.replaceAll("^|(?<=\n)", " " * length)
+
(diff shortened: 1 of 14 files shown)


src/main/scala/temple/utils/StringUtils.scala
```diff
@@ -5,4 +5,12 @@ object StringUtils {
 
   /** Given a string, indent it by a given number of spaces */
   def indent(str: String, length: Int = 2): String = str.replaceAll("^|(?<=\n)", " " * length)
+
+  /** Given a string, convert it to snake case */
+  // https://github.com/lift/framework/blob/f1b450db2dd6a22cf9ffe5576ec34c8e87118319/core/util/src/main/scala/net/liftweb/util/StringHelpers.scala#L91
+  def snakeCase(str: String): String =
+    str
+      .replaceAll("([A-Z]+)([A-Z][a-z])", "$1_$2")
```
c-harding, line 13: <3

src/main/scala/temple/utils/StringUtils.scala
```diff
@@ -5,4 +5,12 @@ object StringUtils {
...
+      .replaceAll("([a-z\\d])([A-Z])", "$1_$2")
```
c-harding, line 14: E2E is now rendered as e2_e, is that desirable?

src/test/scala/temple/utils/StringUtilsTest.scala
```diff
@@ -24,4 +24,34 @@ class StringUtilsTest extends FlatSpec with Matchers {
     indent("abcd\nefg", 1) shouldEqual " abcd\n efg"
     indent("abcd\nefg\n", 1) shouldEqual " abcd\n efg\n "
   }
+
+  behavior of "snakeCase"
```
c-harding, line 28: What should a table CustomerID be rendered as? Can that be enforced with tests?

c-harding (approved):
large_boi

src/test/scala/temple/utils/StringUtilsTest.scala
jaylees14: There was a test where this was checked implicitly, but I've added a more explicit one to make that clear:

https://github.com/TempleEight/temple/pull/83/files#diff-5a2f0a5251b6239394f7b57213f42947R39

Lewiky (approved):

PR #84, opened 2020-02-21
[builder] Rename project argument to files and use global configuration

jaylees14:
There's nothing special about `databaseCreationScripts` - there's no reason this can't be files.
Also added a global project config with defaults for language/database etc to use when one isn't defined.


Lewiky (approved):

willjvsmith (approved):
LGTM 🚀

c-harding (commented):
“But I thought I was special 😭”
— *databaseCreationScripts*, probably

PR #85, opened 2020-02-21
Builder test data

jaylees14:
It might be useful in a future PR to use the `ServiceBlocks` defined in the `DatabaseBuilderIntegrationTestData` for other types of `BuilderIntegregrationTests`... So, let's move them to their own file.


c-harding (approved):
Future us thanks you

willjvsmith (approved):
👏

PR #86, opened 2020-02-21
[builder][docker] Build dockerfile from language

jaylees14:
Let's generate some Dockerfile!
This adds the builder which is language specific (I can't really add anything for the Scala one @xsanda - not sure why it exists?).
The next PR will add the necessary file IO etc.

### Test Plan
Add additional unit tests


c-harding (approved):
Looks good, especially with my updates 😝

Lewiky (approved):
LGTM

PR #89, opened 2020-02-24
[builder][docker] Output Dockerfile for each service, add integration and E2E tests

jaylees14:
- Output generated Dockerfiles
- Add integration and E2E tests


src/main/scala/temple/builder/project/ProjectBuilder.scala
```diff
@@ -28,6 +29,13 @@ object ProjectBuilder {
         }
     }
 
-    Project(databaseCreationScripts)
+    val dockerfiles = templefile.services.zip(80 until Int.MaxValue).map {
```
jaylees14: I'm envisaging the port will eventually come from elsewhere, since it'll also be needed for code generation. For now this seemed the nicest way to express it.

src/e2e/scala/temple/SimpleE2ETest.scala
```diff
@@ -21,15 +21,25 @@ class SimpleE2ETest extends FlatSpec with Matchers {
       ),
     )
 
-    // Only one folder should have been generated
-    Files.list(basePath).count() shouldBe 1
+    // Two folders should have been generated
+    Files.list(basePath).count() shouldBe 2
```
c-harding: Can we check what they are, rather than their count? That way it will be easier to check what the test intended when it doesn’t pass.

c-harding: `Files.list(basePath)` should be a set of x and y

jaylees14: I thought that was sufficiently tested by first checking there was only two, and then checking which folders they were in the following two tests?

jaylees14: But I guess the error if it doesn't succeed will be that 2 != 3, if that's what you mean?

c-harding: Although this does fail when there’s an error, checking the contents of the list should lead to more readable errors (“x missing” rather than 1 != 2)

src/main/scala/temple/builder/project/ProjectBuilder.scala
Lewiky: I suppose this works for now - but definitely in future we probably want a more sophisticated scheme than just incrementing the port - lots of ports we don't want to get confused about as they have conventionally different uses

Lewiky (approved):
LGTM

src/main/scala/temple/builder/project/ProjectBuilder.scala
c-harding: Can we start at a number greater than 1024? That way we won’t run into any conflicts with expected usage. E.g. browsers may try to connect to 443 from 80

jaylees14: Sure - but remember this is a port used in a Docker container. How that's exposed to the host machine is another configuration entirely.

Lewiky (approved):

src/e2e/scala/temple/SimpleE2ETest.scala
```diff
@@ -21,15 +23,24 @@ class SimpleE2ETest extends FlatSpec with Matchers {
...
 
-    // Only one folder should have been generated
-    Files.list(basePath).count() shouldBe 1
-    Files.exists(basePath.resolve("templeuser-db")) shouldBe true
+    // Two folders should have been generated
+    val expectedFolders = Set("templeuser-db", "templeuser").map(dir => basePath.resolve(dir))
+    Files.list(basePath).toScala(Set) shouldBe expectedFolders
 
-    // Only one file should be present in the user-db folder
+    // Only one file should be present in the templeuser-db folder
```
c-harding, line 31: This is the same as above: how about a utility function for checking a folder contains exactly a set of files?

src/e2e/scala/temple/SimpleE2ETest.scala
```diff
@@ -21,15 +23,24 @@ class SimpleE2ETest extends FlatSpec with Matchers {
...
     Files.list(basePath.resolve("templeuser-db")).count() shouldBe 1
     Files.exists(basePath.resolve("templeuser-db").resolve("init.sql")) shouldBe true
 
+    // The content of the templeuser-db/init.sql file should be correct
     val initSql = Files.readString(basePath.resolve("templeuser-db").resolve("init.sql"))
     initSql shouldBe SimpleE2ETestData.createStatement
+
+    // Only one file should be present in the templeuser folder
+    Files.list(basePath.resolve("templeuser")).count() shouldBe 1
+    Files.exists(basePath.resolve("templeuser").resolve("Dockerfile")) shouldBe true
```
c-harding: Ditto

src/e2e/scala/temple/SimpleE2ETest.scala
jaylees14: Not sure what this would look like? I don't like the idea of putting test logic in a utility function.

jaylees14: I'm happy to update the others to follow this new style of checking though

c-harding:
Sorry for being slow to reply, terrible coverage on the train. How about a FileUtils function that takes a folder path and returns a set of contents: directorySet(folder) should be Set(a,b), rather than needing to repeat all that toScala(Set) stuff every time: seems to be a bit too low-level to be repeated so much.

c-harding (approved):
Approving because it looks good on the whole, I trust you to resolve this nicely, and I can’t guarantee I’ll have coverage much in the next hour (cheers [Deutsche Bahn](https://m.youtube.com/watch?v=wXjhszy2f9w))

c-harding (approved):

PR #90, opened 2020-02-24
[go-gen] Add package and import generation

willjvsmith:
* Add file content alias to aid readability
* Add simple package and import generation
  * Based on `user.go` and `match.go`
  * Note, Go uses tabs to indent

This is an iterative PR, a lot of changes will be made as I progress through generating the rest of the files, hence why I don't think we should over-engineer this. Having said that, if you can think of a better way of dealing with the tabs and newlines lmk.

Also note, we discussed using gofmt to handle the formatting for us, but that would require either making it a dependency, or alternatively making Docker a dependency. We decided to use Scala to avoid making Docker a dependency for Swift, so defeats the point.


src/main/scala/temple/generate/language/service/go/GoServiceGenerator.scala
```diff
@@ -0,0 +1,56 @@
...
+    |	"net/http"""".stripMargin + "\n\n")
+
+    if (comms) {
+      sb.append(tabIndent(s"""${serviceName}Comm "${module}/comm"""") + "\n")
+    }
+
+    sb.append(tabIndent(s"""${serviceName}DAO "${module}/dao"""") + "\n")
+    sb.append(tabIndent(s""""${module}/util"""") + "\n")
+
+    sb.append("""	valid "github.com/asaskevich/govalidator"
```
c-harding: tabIndent plz

src/main/scala/temple/generate/language/service/go/GoServiceGenerator.scala
```diff
@@ -0,0 +1,56 @@
...
+import temple.utils.StringUtils.tabIndent
+
+object GoServiceGenerator extends ServiceGenerator {
+
+  private def generatePackage(packageName: String): String = s"package $packageName\n\n"
+
+  private def generateImports(serviceName: String, module: String, comms: Boolean): String = {
+    val sb = new StringBuilder
+
+    sb.append("""import (
```
c-harding: Is there any way we can avoid the parens being separated? Could these strings be wrapped in quotes, separated by new lines, and indented, then we call codeParens on them?

src/main/scala/temple/utils/StringUtils.scala
```diff
@@ -6,6 +6,9 @@ object StringUtils {
   /** Given a string, indent it by a given number of spaces */
   def indent(str: String, length: Int = 2): String = str.replaceAll("^|(?<=\n)", " " * length)
 
+  /** Given a string, indent it by a given number of tabs */
+  def tabIndent(str: String, length: Int = 1): String = str.replaceAll("^|(?<=\n)", "	" * length)
```
c-harding: Does `\t` work instead of a literal tab in the source code?

src/test/scala/temple/builder/project/ProjectBuilderTest.scala
```diff
@@ -1,7 +1,7 @@
 package temple.builder.project
 
 import org.scalatest.{FlatSpec, Matchers}
-import temple.builder.project.Project.File
+import temple.utils.FileUtils.{File}
```
c-harding: unnecessary {}, might be worth running Optimize Imports

src/main/scala/temple/generate/language/service/ServiceGenerator.scala
```diff
@@ -0,0 +1,8 @@
+package temple.generate.language.service
+
+import temple.generate.language.service.adt._
+import temple.utils.FileUtils.{File, FileContent}
+
+trait ServiceGenerator {
+  def generate(serviceRoot: ServiceRoot): Map[File, FileContent]
+}
```
jaylees14: Some docs here would be good

src/main/scala/temple/generate/language/service/go/GoServiceGenerator.scala
```diff
@@ -0,0 +1,56 @@
...
+import temple.generate.language.service.ServiceGenerator
+import temple.generate.language.service.adt._
+import temple.utils.FileUtils.{File, FileContent}
+import temple.utils.StringUtils.tabIndent
+
+object GoServiceGenerator extends ServiceGenerator {
+
+  private def generatePackage(packageName: String): String = s"package $packageName\n\n"
+
+  private def generateImports(serviceName: String, module: String, comms: Boolean): String = {
```
jaylees14, line 16: To the uninitiated, it might not be clear what `comms` means. Might be worth adding a comment

src/main/scala/temple/builder/project/ProjectBuilder.scala
```diff
@@ -4,7 +4,7 @@ import temple.DSL.semantics.Metadata.Database
 import temple.DSL.semantics.Metadata.Database.Postgres
 import temple.DSL.semantics.Templefile
 import temple.builder.DatabaseBuilder
-import temple.builder.project.Project.File
+import temple.utils.FileUtils.{File}
```
jaylees14: Think we can drop the `{}` here too

src/main/scala/temple/generate/language/service/go/GoServiceGenerator.scala
```diff
@@ -2,22 +2,26 @@ package temple.generate.language.service.go
...
-    |	"fmt"
-    |	"log"
-    |	"net/http"""".stripMargin + "\n\n")
+    sb.append("import (" + "\n")
+    sb.append(tabIndent(""""encoding/json"""") + "\n")
+    sb.append(tabIndent(""""flag"""") + "\n")
+    sb.append(tabIndent(""""fmt"""") + "\n")
+    sb.append(tabIndent(""""log"""") + "\n")
+    sb.append(tabIndent(""""net/http"""") + "\n")
+    sb.append("\n")
```
c-harding: Now factor out all the repetition (including below) and I’ll be very happy

c-harding: Something like this? n.b. `spaced` will need to be adapted to support tabs, and named imports will need to be added (tuple with optional string?). This could be bundled up into a go-specific set of codegen utility functions; I’m sure this isn’t the only place imports will be desired.
```scala
codeParens.spaced(Seq("encoding/json", ???, "net/http").map(enquote))
```

willjvsmith:
@xsanda Please see the unit test errors, this is caused by the logic/regex in the `apply` function of `mkCode` in `CodeTerm.scala`.

If we add single and double quotes to the regexes, a different Postgres test fails, where a string with single quotes already in it is passed in, e.g. `'Will'`. See `PostgresGeneratorTestData.scala:232`.

willjvsmith:
> can a `ListBuffer` be used here (with `+=`/`++=` appending), to avoid the repetition of `customImports = customImports +`?

Out of curiosity, why a `ListBuffer` over a mutable `Seq`?

c-harding:
> > can a `ListBuffer` be used here (with `+=`/`++=` appending), to avoid the repetition of `customImports = customImports +`?
> 
> Out of curiosity, why a `ListBuffer` over a mutable `Seq`?

A `ListBuffer` is an instance of `mutable.Seq`, specifically one with O(1) `append` and `toSeq`

willjvsmith:
I will update to using `CodeWrap.spaced` when #94 is ready and merged, and the issue causing the tests to fail is resolved.

c-harding:
> @xsanda Please see the unit test errors, this is caused by the logic/regex in the `apply` function of `mkCode` in `CodeTerm.scala`.
> 
> If we add single and double quotes to the regexes, a different Postgres test fails, where a string with single quotes already in it is passed in, e.g. `'Will'`. See `PostgresGeneratorTestData.scala:232`.

You’re right, `CodeWrap` is not right for spaces as it takes code, not strings, as contents

src/main/scala/temple/generate/language/service/go/GoServiceGenerator.scala
```diff
@@ -0,0 +1,63 @@
...
+    customImports = customImports.map(tabIndent(_)).map(_ + "\n")
+
+    sb.append(codeWrap.parens.block(Seq(standardImports.mkString, "\n", customImports.mkString).mkString))
+    sb.append("\n")
+
+    sb.toString
+  }
+
+  override def generate(serviceRoot: ServiceRoot): Map[File, FileContent] = {
+    val usesComms = serviceRoot.comms.length > 0
```
c-harding: IntelliJ says this should be replaced with `serviceRoot.comms.nonEmpty`

src/main/scala/temple/generate/language/service/go/GoServiceGenerator.scala
```diff
@@ -0,0 +1,63 @@
...
+    val serviceString = generatePackage("main") + generateImports(
+        serviceRoot.name,
+        serviceRoot.module,
+        usesComms,
+      )
+    Map(File(serviceRoot.name, s"${serviceRoot.name}.go") -> serviceString)
+    /**
```
c-harding: `/**` is for documentation, `/*` is for comments

jaylees14 (approved):
LGTM 👍

Lewiky (approved):
LGTM

PR #91, opened 2020-02-24
[test][golang] Add integration test support for Golang

jaylees14:
This adds support for validation of Golang files - should hopefully be useful for @smithwjv to do some TDD 😉.

Couple of changes:
- Essentially we need the same configuration as Hadolint (invoke a binary on HTTP request), so I've bundled these in one big container, so we can have a single deployment.
- This means that it has become a more general `shell2http` container, rather than just Hadolint
- However, this had the side effect that two test classes can't use the same underlying container (since it tries to spin up two containers of the same image but on the same port) 
- I've done some rearchitecting of how we construct each of these containers so that they both run, but on separate ports, and this is provided as a constructor argument.


src/it/scala/temple/generate/language/service/GoGeneratorIntegrationTest.scala
```diff
@@ -0,0 +1,29 @@
+package temple.generate.language.service
+
+import org.scalatest.{BeforeAndAfter, Matchers}
+import temple.containers.GolangSpec
+import temple.generate.docker.ast.DockerfileRoot
+import temple.generate.docker.ast.Statement._
+
+class GoGeneratorIntegrationTest extends GolangSpec with Matchers with BeforeAndAfter {
+
+  behavior of "GolangValidator"
```
smithwjv: \n

willjvsmith, line 10: \n

willjvsmith (approved):
LGTM

src/it/scala/temple/generate/language/service/GoGeneratorIntegrationTest.scala
jaylees14: We haven't been very consistent with this... See:

https://github.com/TempleEight/temple/blob/07b33ef2cb0023fe2bcf3019b6b2289f1e3071c4/src/it/scala/temple/generate/database/PostgresGeneratorIntegrationTest.scala#L24-L25

https://github.com/TempleEight/temple/blob/07b33ef2cb0023fe2bcf3019b6b2289f1e3071c4/src/test/scala/temple/generate/database/PostgresGeneratorTest.scala#L10-L12

Which do we prefer? I'll fix them all in a follow up.

src/it/scala/temple/containers/DockerShell2HttpService.scala
```diff
@@ -0,0 +1,34 @@
+package temple.containers
+
+import com.whisk.docker.{DockerCommandExecutor, DockerContainer, DockerContainerState, DockerKit, DockerReadyChecker}
+import org.scalatest.FlatSpec
+import scalaj.http.Http
+
+import scala.concurrent.{ExecutionContext, Future}
+import scala.concurrent.duration._
+
+/** DockerShell2HttpService configures a docker container for our custom Shell2Http container */
```
Lewiky, line 10: Love the Go style docstrings here ;)

src/it/scala/temple/generate/language/service/GoGeneratorIntegrationTest.scala
Lewiky: I like the newline personally

Lewiky (approved):

PR #92, opened 2020-02-24
[test] Use consistent newlines between behavior and tests

jaylees14:
**Big** changes. From #91, use consistent spacing everywhere in tests.


willjvsmith (approved):

c-harding (approved):
This is what I like to see

PR #93, opened 2020-02-24
[go-gen] Generalise codeparens

willjvsmith:
* Generalise `codeParens` to `codeWrap`

I've added this to make my Go generation nicer

### DIFF
--- a/src/main/scala/temple/generate/utils/CodeTerm.scala
+++ b/src/main/scala/temple/generate/utils/CodeTerm.scala
@@ -87,7 +87,7 @@ object CodeTerm {
     def stmt(string: CodeTerm*): String = mkCode(string, ";")
   }
 
-  sealed class codeParens private (start: String, end: String) {
+  sealed class codeWrap private (start: String, end: String) {
 
     /** Wrap a code snippet in parentheses */
     def apply(string: CodeTerm*): String = mkCode(start, string, end)
@@ -96,8 +96,22 @@ object CodeTerm {
     def spaced(string: CodeTerm*): String = mkCode(start, "\n", indent(mkCode(string)), "\n", end)
   }
 
-  object codeParens extends codeParens("(", ")") {
-    object curly  extends codeParens("{", "}")
-    object square extends codeParens("[", "]")
+  object codeWrap {
+
+    object parens extends codeWrap("(", ")") {
+      object block extends codeWrap("(\n", ")\n")
+    }
+
+    object curly extends codeWrap("{", "}") {
+      object block extends codeWrap("{\n", "}\n")
+    }
+
+    object square extends codeWrap("[", "]") {
+      object block extends codeWrap("[\n", "]\n")
+    }
+
+    object singleQuotes extends codeWrap("'", "'")
+    object doubleQuotes extends codeWrap("\"", "\"")
+    object españolQue   extends codeWrap("¿", "?")
   }
 }
--- a/src/main/scala/temple/DSL/syntax/Entry.scala
+++ b/src/main/scala/temple/DSL/syntax/Entry.scala
@@ -1,7 +1,7 @@
 package temple.DSL.syntax
 
 import temple.DSL.syntax.Arg.ListArg
-import temple.generate.utils.CodeTerm.{codeParens, mkCode}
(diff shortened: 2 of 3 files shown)


jaylees14 (approved):
LGTM

src/main/scala/temple/generate/utils/CodeTerm.scala
```diff
@@ -96,8 +96,22 @@ object CodeTerm {
...
+      object block extends codeWrap("{\n", "}\n")
+    }
+
+    object square extends codeWrap("[", "]") {
+      object block extends codeWrap("[\n", "]\n")
+    }
+
+    object singleQuotes extends codeWrap("'", "'")
+    object doubleQuotes extends codeWrap("\"", "\"")
+    object españolQue   extends codeWrap("¿", "?")
```
Lewiky, line 115: This is a serious project, we have no time for spanish tomfoolery

Lewiky (approved):

src/main/scala/temple/generate/utils/CodeTerm.scala
```diff
@@ -96,8 +96,22 @@ object CodeTerm {
     def spaced(string: CodeTerm*): String = mkCode(start, "\n", indent(mkCode(string)), "\n", end)
   }
 
-  object codeParens extends codeParens("(", ")") {
-    object curly  extends codeParens("{", "}")
-    object square extends codeParens("[", "]")
+  object codeWrap {
+
+    object parens extends codeWrap("(", ")") {
+      object block extends codeWrap("(\n", ")\n")
```
c-harding, line 102: That should be `\n)`, shouldn’t it?

src/main/scala/temple/generate/utils/CodeTerm.scala
```diff
@@ -96,8 +96,22 @@ object CodeTerm {
     def spaced(string: CodeTerm*): String = mkCode(start, "\n", indent(mkCode(string)), "\n", end)
   }
 
-  object codeParens extends codeParens("(", ")") {
-    object curly  extends codeParens("{", "}")
-    object square extends codeParens("[", "]")
+  object codeWrap {
```
c-harding, line 99: As this is no longer callable, can `codeWrap` (both the class and the companion object) have a leading capital please?

src/main/scala/temple/generate/utils/CodeTerm.scala
c-harding: Is this variable naming legal?

src/main/scala/temple/generate/utils/CodeTerm.scala
```diff
@@ -96,8 +96,22 @@ object CodeTerm {
...
+
+    object curly extends codeWrap("{", "}") {
+      object block extends codeWrap("{\n", "}\n")
+    }
+
+    object square extends codeWrap("[", "]") {
+      object block extends codeWrap("[\n", "]\n")
+    }
+
+    object singleQuotes extends codeWrap("'", "'")
```
c-harding, line 113: This looks like it was my error from before, but can this be converted to a `val` rather than an object?

src/main/scala/temple/generate/utils/CodeTerm.scala
```diff
@@ -96,8 +96,22 @@ object CodeTerm {
...
+    object parens extends codeWrap("(", ")") {
+      object block extends codeWrap("(\n", ")\n")
+    }
+
+    object curly extends codeWrap("{", "}") {
+      object block extends codeWrap("{\n", "}\n")
+    }
+
+    object square extends codeWrap("[", "]") {
+      object block extends codeWrap("[\n", "]\n")
```
c-harding, line 110:
```scala
def block: CodeWrap = new CodeWrap(s"$start\n",s"\n$end")
```
On the class, rather than repeating per object. If you don’t want to enable this for quotes, then make a subclass CodeBlockWrap or similar. Then square, curly etc can be made vals rather than objects

c-harding:
Damn it, was too slow. Will add a PR

src/main/scala/temple/generate/utils/CodeTerm.scala
c-harding: This already exists, it’s CodeWrap#spaced

PR #94, opened 2020-02-24
[go-gen] Simplify CodeWrap

c-harding:
Jumping on the back of #93 with some simplifications. The block methods have been removed as they are the same as `CodeWrap#spaced`.

### DIFF
--- a/src/main/scala/temple/generate/utils/CodeTerm.scala
+++ b/src/main/scala/temple/generate/utils/CodeTerm.scala
@@ -87,7 +87,7 @@ object CodeTerm {
     def stmt(string: CodeTerm*): String = mkCode(string, ";")
   }
 
-  sealed class codeWrap private (start: String, end: String) {
+  sealed class CodeWrap private (start: String, end: String) {
 
     /** Wrap a code snippet in parentheses */
     def apply(string: CodeTerm*): String = mkCode(start, string, end)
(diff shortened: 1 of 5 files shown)


src/main/scala/temple/generate/utils/CodeTerm.scala
```diff
@@ -96,22 +96,11 @@ object CodeTerm {
...
-      object block extends codeWrap("{\n", "}\n")
-    }
-
-    object square extends codeWrap("[", "]") {
-      object block extends codeWrap("[\n", "]\n")
-    }
-
-    object singleQuotes extends codeWrap("'", "'")
-    object doubleQuotes extends codeWrap("\"", "\"")
-    object españolQue   extends codeWrap("¿", "?")
```
willjvsmith, line 115: Not accepting this PR until this is back in

willjvsmith (commented):
The blocks aren't the same as `CodeWrap.spaced` as I added the "\n" after the closing brace, also consider this case.

```
import (
  "fmt"

  "gitHub.com..."
)
```

The blank line between `"fmt"` and `"gitHub.com..."` should not be indented if we want to match the output of `gofmt`, `CodeWrap.spaced` does not do this.

src/main/scala/temple/utils/StringUtils.scala
```diff
@@ -3,8 +3,11 @@ package temple.utils
 /** Utility functions useful for performing operations on strings */
 object StringUtils {
 
+  /** Match the position at the start of a line, excluding blank lines */
+  private val startOfLinePattern = """^|(?<=\n)(?!\n)(?!$)""".r
```
willjvsmith: @jaylees14 is gonna love this

willjvsmith (approved):

src/main/scala/temple/utils/StringUtils.scala
c-harding: Does he love its replacement?

jaylees14 (approved):
Much more readable 👍

PR #95, opened 2020-02-24
[go-gen] Enquote

c-harding:
Add function to wrap in quotes

### DIFF
--- a/src/main/scala/temple/generate/utils/CodeTerm.scala
+++ b/src/main/scala/temple/generate/utils/CodeTerm.scala
@@ -106,7 +106,7 @@ object CodeTerm {
       object block extends codeWrap("{\n", "}\n")
     }
 
-    object square extends codeWrap("[" ush, "]") {
+    object square extends codeWrap("[", "]") {
       object block extends codeWrap("[\n", "]\n")
     }
   }
--- a/src/main/scala/temple/generate/utils/CodeTerm.scala
+++ b/src/main/scala/temple/generate/utils/CodeTerm.scala
@@ -102,13 +102,4 @@ object CodeTerm {
     val square = new CodeWrap("[", "]")
   }
 
-  type StringWrap = String => String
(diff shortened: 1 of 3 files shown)


src/main/scala/temple/generate/utils/CodeTerm.scala
```diff
@@ -106,12 +106,16 @@ object CodeTerm {
       object block extends codeWrap("{\n", "}\n")
     }
 
-    object square extends codeWrap("[", "]") {
+    object square extends codeWrap("[" ush, "]") {
```
willjvsmith: what dis

willjvsmith:
I think a better solution would be a generalised StringUtils function, that defaults to double quotes, but allows you to supply the string to wrap with, rather than only allowing JSON-encoding a string. I don't think `CodeTerm.scala` is the right place for it.

willjvsmith:
I'll take over this PR tomorrow when we're inside working hours 😉

src/main/scala/temple/generate/utils/CodeTerm.scala
```diff
@@ -109,9 +109,15 @@ object CodeTerm {
     object square extends codeWrap("[", "]") {
       object block extends codeWrap("[\n", "]\n")
     }
-
-    object singleQuotes extends codeWrap("'", "'")
-    object doubleQuotes extends codeWrap("\"", "\"")
-    object españolQue   extends codeWrap("¿", "?")
```
Lewiky: ¿Por qué no hay amor para los españoles?

Lewiky (approved):

src/main/scala/temple/generate/utils/CodeTerm.scala
c-harding: Will está a cargo ahora, puede agregarlo de nuevo si quiere

willjvsmith: español está de vuelta en el menú

PR #96, opened 2020-02-26
[go-gen] Add main generation

willjvsmith:
* Generate main function
* Update tests

### DIFF
--- a/src/main/scala/temple/generate/language/service/adt/Endpoint.scala
+++ b/src/main/scala/temple/generate/language/service/adt/Endpoint.scala
@@ -3,9 +3,9 @@ package temple.generate.language.service.adt
 sealed trait Endpoint
 
 object Endpoint {
-  case object All    extends Endpoint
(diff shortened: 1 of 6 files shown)


c-harding (approved):
Very nice! Is it possible to add some integration tests to check that the stuff it outputs compiles? (Or that the code in TestData compiles, as we’re asserting that they’re identical)

willjvsmith:
> Very nice! Is it possible to add some integration tests to check that the stuff it outputs compiles? (Or that the code in TestData compiles, as we’re asserting that they’re identical)

At the moment the files aren't complete so they won't compile anyway, but I'll definitely add integration tests once everything is generated!

jaylees14 (approved):
LGTM - just echoing what Charlie said, if you could make the produced code compile within the next PR or so, then add correct functionality (e.g. add the handlers with empty bodies), that'll be smashing 👍

PR #97, opened 2020-02-26
[go-gen] Add global, empty handler, jsonMiddleware and errors.go generation

willjvsmith:
* Add global generation
* Add generation of empty handlers, to speed up progress to a point where the code generated can be compiled
* Add jsonMiddleware function from file
* Reorder `dao` and `comms` imports to be consistent
* Add generation of the `errors.go` `dao` file

Lmk if there's a better way of giving a string representation of the endpoint enum


src/main/scala/temple/generate/language/service/go/GoServiceGenerator.scala
```diff
@@ -96,22 +107,71 @@ object GoServiceGenerator extends ServiceGenerator {
     sb.toString
   }
 
+  private def generateJsonMiddleware(): String =
+    FileUtils.readFile("src/main/scala/temple/generate/language/service/go/genFiles/JsonMiddleware.go")
+
+  private def generateHandler(serviceName: String, endpoint: Endpoint): String =
+    s"func ${serviceName + endpoint}Handler(w http.ResponseWriter, r *http.Request) {}\n\n"
```
jaylees14: Does this still work when we've made a `.jar`? Does it bundle the `.go` files up?

src/main/scala/temple/generate/language/service/go/GoServiceGenerator.scala
```diff
@@ -96,22 +107,71 @@ object GoServiceGenerator extends ServiceGenerator {
...
+    }
+    if (endpoints.contains(Endpoint.Read)) {
+      sb.append(generateHandler(serviceName, Endpoint.Read))
+    }
+    if (endpoints.contains(Endpoint.Update)) {
+      sb.append(generateHandler(serviceName, Endpoint.Update))
+    }
+    if (endpoints.contains(Endpoint.Delete)) {
+      sb.append(generateHandler(serviceName, Endpoint.Delete))
+    }
```
jaylees14: I see what you're doing here... I'm wondering if we could define a `sort` for these and just call `generateHandler` in a loop?

jaylees14 (commented):
Few things I'm not too sure about. Also wondering if rather than adding a new line after each "chunk" , you form a list and then reduce them by separating with a new line (i.e `Seq(imports, globals).mkString("\n")`)? Not sure if I'm missing an edge case with that

src/main/scala/temple/generate/language/service/go/GoServiceGenerator.scala
```diff
@@ -96,22 +106,71 @@ object GoServiceGenerator extends ServiceGenerator {
     sb.toString
   }
 
+  private def generateJsonMiddleware(): String =
+    readFile("src/main/scala/temple/generate/language/service/go/genFiles/JsonMiddleware.go")
```
jaylees14: Does this still work when we've made a `.jar`? Does it bundle the `.go` files up?

jaylees14: Think I prefer the explicit `FileUtils`

src/main/scala/temple/generate/language/service/go/GoServiceGenerator.scala
c-harding: One simple way, if `values` is defined as a list in Endpoint’s companion object
```scala
for (endpoint <- Endpoint.values if endpoints contains endpoint) {
  sb.append(generateHandler(serviceName, Endpoint.Delete))
}
```

src/main/scala/temple/generate/language/service/go/GoServiceGenerator.scala
```diff
@@ -96,22 +106,71 @@ object GoServiceGenerator extends ServiceGenerator {
...
+  private def generateErrors(serviceName: String): String =
+    Seq(
+      "package dao",
+      "",
+      """import "fmt"""",
+      "",
+      s"// Err${serviceName.capitalize}NotFound is returned when a ${serviceName} for the provided ID was not found",
+      s"type Err${serviceName.capitalize}NotFound int64",
+      "",
+      s"func (e Err${serviceName.capitalize}NotFound) Error() string " + CodeWrap.curly
```
c-harding: personally this feels like a bit of an odd mix and match between interpolation and concatenation

src/main/scala/temple/generate/language/service/go/GoServiceGenerator.scala
jaylees14: Does `.values` preserve an ordering?

src/main/scala/temple/generate/language/service/go/GoServiceGenerator.scala
```diff
@@ -96,22 +106,71 @@ object GoServiceGenerator extends ServiceGenerator {
...
         serviceRoot.name,
         usesComms,
         serviceRoot.endpoints,
         serviceRoot.port,
-      )
-    Map(File(serviceRoot.name, s"${serviceRoot.name}.go") -> serviceString)
+      ) + generateJsonMiddleware() + generateHandlers(
+        serviceRoot.name,
+        serviceRoot.endpoints,
+      ).stripLineEnd
```
c-harding: A little confused as to why we strip the trailing newline here: should these functions return trailing newlines at all?

src/main/scala/temple/generate/language/service/go/GoServiceGenerator.scala
```diff
@@ -96,22 +106,71 @@ object GoServiceGenerator extends ServiceGenerator {
     sb.toString
   }
 
+  private def generateJsonMiddleware(): String =
+    readFile("src/main/scala/temple/generate/language/service/go/genFiles/JsonMiddleware.go")
+
+  private def generateHandler(serviceName: String, endpoint: Endpoint): String =
+    s"func ${serviceName + endpoint}Handler(w http.ResponseWriter, r *http.Request) {}\n\n"
```
c-harding: Is this clearer than $serviceName$endpoint?

src/main/scala/temple/generate/language/service/go/GoServiceGenerator.scala
c-harding: It would need to be defined manually, so just make it a `Seq`

willjvsmith: > One simple way, if `values` is defined as a list in Endpoint’s companion object
> 
> ```scala
> for (endpoint <- Endpoint.values if endpoints contains endpoint) {
>   sb.append(generateHandler(serviceName, Endpoint.Delete))
> }
> ```

Nice solution, didn't know about this filtering notation

src/main/scala/temple/generate/language/service/go/GoServiceGenerator.scala
willjvsmith: It's certainly no less clear

src/main/scala/temple/generate/language/service/go/GoServiceGenerator.scala
willjvsmith: > Does this still work when we've made a `.jar`? Does it bundle the `.go` files up?

@Lewiky

If we find out it doesn't, it won't be hard to change

src/main/scala/temple/generate/language/service/go/GoServiceGenerator.scala
willjvsmith: I agree, though I'm not sure I much prefer what I've changed it to, thoughts?

src/main/scala/temple/generate/language/service/adt/Endpoint.scala
```diff
@@ -3,9 +3,26 @@ package temple.generate.language.service.adt
 sealed trait Endpoint
 
 object Endpoint {
-  case object ReadAll extends Endpoint
-  case object Create  extends Endpoint
-  case object Read    extends Endpoint
-  case object Update  extends Endpoint
-  case object Delete  extends Endpoint
+
+  case object ReadAll extends Endpoint {
```
c-harding: I’d be tempted to refactor this to
```scala
sealed abstract class Endpoint(verb: String)
case object ReadAll extends Endpoint("List")
```
Firstly so it’s less verbose, and secondly because it makes it slightly clearer what the string represents

src/main/scala/temple/generate/language/service/go/GoServiceGenerator.scala
```diff
@@ -92,26 +102,74 @@ object GoServiceGenerator extends ServiceGenerator {
...
-  override def generate(serviceRoot: ServiceRoot): Map[File, FileContent] = {
+  private def generateJsonMiddleware(): String =
+    FileUtils.readFile("src/main/scala/temple/generate/language/service/go/genFiles/JsonMiddleware.go")
+
+  private def generateHandler(serviceName: String, endpoint: Endpoint): String =
+    s"func $serviceName${endpoint}Handler(w http.ResponseWriter, r *http.Request) {}\n"
+
+  private def generateHandlers(serviceName: String, endpoints: Set[Endpoint]): String = {
+    val handlers: ListBuffer[String] = ListBuffer.empty
+    for (endpoint <- Endpoint.values if endpoints.contains(endpoint)) {
```
c-harding:
```scala
val handlers = for (...) yield generateHandler(serviceName, endpoint)
```

c-harding (approved):
Looks like it does the job

src/main/scala/temple/generate/language/service/go/GoServiceGenerator.scala
willjvsmith: good catch

src/main/scala/temple/generate/language/service/go/GoServiceGenerator.scala
c-harding: That’s still a mix of concat and interpolation, just with slightly more long-winded concatenation

c-harding: Maybe reproduces the structure slightly artificially in the code, but how does this look?
```scala
errors +=
  s"func (e Err${serviceName.capitalize}NotFound) Error() string ${CodeWrap.curly.tabbed(
    s"""return fmt.Sprintf("${serviceName} not found with ID %d", e)""",
  )}"
```

src/main/scala/temple/generate/language/service/adt/Endpoint.scala
willjvsmith: > I’d be tempted to refactor this to
> 
> ```scala
> sealed abstract class Endpoint(verb: String)
> case object ReadAll extends Endpoint("List")
> ```
> 
> Firstly so it’s less verbose, and secondly because it makes it slightly clearer what the string represents

This needs `val` in front of verb to be usable

c-harding: yep, my bad

src/main/scala/temple/generate/language/service/go/GoServiceGenerator.scala
willjvsmith: If you're happy with that lol

c-harding: I feel it’s nicer than the weird way that `CodeWrap.curly.tabbed` was being line-broken before, but up to you

jaylees14 (approved):
LGTM

PR #98, opened 2020-02-26
[code-gen] Generate line-separated code

c-harding:
The idea here is that `CodeWrap.parens.tabbed` can take `CodeTerms`, meaning strings, `Seq`s and `Option`s thereof.

This means a function like `generateImports` can go from
```scala
  private def generateImports(serviceName: String, module: String, usesComms: Boolean): String = {
    val sb = new StringBuilder
    sb.append("import ")

    val standardImports = Seq("encoding/json", "flag", "fmt", "log", "net/http")
      .map(doubleQuote)
      .mkString("\n")

    var customImportsBuffer: ListBuffer[String] = ListBuffer.empty
    if (usesComms) {
      customImportsBuffer += s"${serviceName}Comm ${doubleQuote(s"${module}/comm")}"
    }
    customImportsBuffer ++= Seq(
      s"${serviceName}DAO ${doubleQuote(s"${module}/dao")}",
      doubleQuote(s"${module}/util"),
      s"valid ${doubleQuote("github.com/asaskevich/govalidator")}",
      doubleQuote("github.com/gorilla/mux"),
    )
    val customImports = customImportsBuffer.mkString("\n")

    sb.append(CodeWrap.parens.tabbed(standardImports + "\n\n" + customImports))
    sb.append("\n\n")
    sb.toString
  }
```
to
```scala
  import scala.Option.when

  private def generateImports(serviceName: String, module: String, usesComms: Boolean): String = {
    val standardImports = Seq("encoding/json", "flag", "fmt", "log", "net/http").map(doubleQuote)

    val customImports = Seq[CodeTerm](
      when(usesComms) { s"${serviceName}Comm ${doubleQuote(s"${module}/comm")}" },
      s"${serviceName}DAO ${doubleQuote(s"${module}/dao")}",
      doubleQuote(s"${module}/util"),
      s"valid ${doubleQuote("github.com/asaskevich/govalidator")}",
      doubleQuote("github.com/gorilla/mux"),
      s"${serviceName}DAO ${doubleQuote(s"$module/dao")}",
    )

    mkCode("import", CodeWrap.parens.tabbed(standardImports, "", customImports))
  }
```

N.b. this is not exactly a slot-in replacement: it doesn’t include the trailing newlines. This means it would become the job of the caller to add newlines, e.g.:

```scala
mkCode.lines(
  generatePackage("main"),
  "",
  generateImports(???)
)
```


jaylees14 (approved):
I'm really struggling to follow any changes that are made in this file. There are so many levels of indirection going on here, for what seems to essentially be separating some strings with commas and newlines. I'm approving it because the example you give looks sensible, but we should be careful about over engineering things like this in the future, it makes it so much harder to review and even harder for others to make changes to it. I think time would be better spent focussing on building new features.

willjvsmith (approved):

PR #99, opened 2020-02-27
[go-gen] Add config, util and placeholder dao and comm generation

willjvsmith:
* Add `config.go` generation from file
* Add `util.go` generation from file
* Add simple `dao.go` without methods
* Add simple `handler.go` without methods

This PR looks big but is made up mostly of copying in files from `spec-golang`! Though I have made small changes to fix typos etc, expect a PR over in `spec-golang` to make them identical in the coming days.

This hopefully gets the generated Go to a point where it can be compiled, next PR will be adding integration tests to test this.

Once compiling I will refactor this in a separate PR.


Lewiky:
In regards to having static go files in the project, in order for resources to be assembled into the jar correctly, they need to be placed in `src/assembly/resources`, as per https://github.com/sbt/sbt-assembly/issues/258

jaylees14 (approved):
LGTM - I think we're at a point now where we can start generalising the code a little - couple of utility functions for making an import statement given a list of imports or a struct given a name and body?

Lewiky (approved):

PR #101, opened 2020-02-28
[go-gen] Fix utils->util, and match gofmt output

willjvsmith:
* Updates generated code with fixes for things found while tidying up the sample service


Lewiky (approved):
KATAMARU DAMASUUUUUUUU

jaylees14 (approved):

PR #103, opened 2020-03-02
[it][test] Run go mod tidy prior to building the project

jaylees14:
Currently dependency pulling causes integration tests to fail, so this does that before building the project.

NB: Can't use `&&`, since `go mod tidy` will fail if it can't find a project, but this isn't echo'd back to the user since output is redirected.


Lewiky (approved):

willjvsmith (approved):
I trust this works, will see shortly!

c-harding (approved):

PR #105, opened 2020-03-02
[go-gen] Add integration tests for compiling generated Go

willjvsmith:
* Adds integration tests for compiling generated Go service code
* Removes unused imports from generated code, will need to be added back in when generation of handlers and DAO code is added
* Replaces `empty` with `""` in integration tests so the compilation errors are displayed


c-harding (approved):

jaylees14 (approved):
LGTM

Issue #107, opened 2020-03-05
Generate OpenAPI spec for generating clients


c-harding:
The file to generate is here:

https://github.com/TempleEight/spec-golang/blob/develop/api/spec-golang.openapi.yaml

willjvsmith:
I think this was mistakenly put here instead of on Monday.

PR #108, opened 2020-03-05
[openapi-gen] Lift Endpoint to cover all generation

c-harding:
- Move Endpoint from `temple.generate.language.service.adt` to `temple.generate`
- Move `temple.generate.language.service` to `temple.generate.service`, as it does not share anything with the client code, so nothing else would go in `temple.generate.language`
- Move `Endpoint#verb` to a private method `ServiceGenerator.verb`, as it is specific to service code.

This is in preparation for fixing #107


src/main/scala/temple/generate/service/go/GoServiceGenerator.scala
```diff
@@ -212,4 +212,12 @@ object GoServiceGenerator extends ServiceGenerator {
...
     )).map { case (path, contents) => path -> (contents + "\n") }
   }
+
+  private def verb(endpoint: Endpoint): String = endpoint match {
+    case Endpoint.ReadAll => "List"
+    case Endpoint.Create  => "Create"
+    case Endpoint.Read    => "Read"
+    case Endpoint.Update  => "Update"
+    case Endpoint.Delete  => "Delete"
+  }
```
c-harding: should this instead be placed in `ServiceGenerator`? Is it Go-specific?

willjvsmith: 🚚 -> Endpoint

willjvsmith (approved):

src/main/scala/temple/generate/service/go/GoServiceGenerator.scala
jaylees14: Don't think so, happy for it to be moved

jaylees14 (approved):
LGTM

PR #110, opened 2020-03-05
[go-gen] Refactor go code generation by package

willjvsmith:
* Replaces `Seqs` with `mkCode.lines` in the `generateMain` function in `GoServiceMainGenerator`
* Refactors the code generation by moving each go package to be generated into their own files

No new code generation here

### DIFF
--- a/src/main/scala/temple/generate/service/go/GoServiceMainGenerator.scala
+++ b/src/main/scala/temple/generate/service/go/GoServiceMainGenerator.scala
@@ -1,0 +1,102 @@
+package temple.generate.service.go
+
+import temple.generate.Endpoint
+import temple.generate.service.ServiceGenerator
+import temple.generate.utils.CodeTerm
+import temple.generate.utils.CodeTerm.{CodeWrap, mkCode}
+import temple.utils.FileUtils
+import temple.utils.StringUtils.doubleQuote
+
+import scala.Option.when
+
+object GoServiceMainGenerator {
+
+  /** Given a service name, module name and whether the service uses inter-service communication, return the import
+    * block */
+  private[go] def generateImports(serviceName: String, module: String, usesComms: Boolean): String = {
+    val standardImports = Seq("flag", "log", "net/http").map(doubleQuote)
+
+    val customImports = Seq[CodeTerm](
+      when(usesComms) { s"${serviceName}Comm ${doubleQuote(s"$module/comm")}" },
+      s"${serviceName}DAO ${doubleQuote(s"$module/dao")}",
+      doubleQuote(s"$module/util"),
+      s"valid ${doubleQuote("github.com/asaskevich/govalidator")}",
+      doubleQuote("github.com/gorilla/mux"),
+    )
+
+    mkCode("import", CodeWrap.parens.tabbed(standardImports, "", customImports))
+  }
+
+  /** Given a service name and whether the service uses inter-service communication, return global statements */
+  private[go] def generateGlobals(serviceName: String, usesComms: Boolean): String =
+    mkCode.lines(
+      s"var dao ${serviceName}DAO.DAO",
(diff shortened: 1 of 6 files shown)


jaylees14 (approved):

src/main/scala/temple/generate/service/go/GoServiceMainGenerator.scala
```diff
@@ -0,0 +1,102 @@
...
+    mkCode.lines(
+      s"var dao ${serviceName}DAO.DAO",
+      when(usesComms) { s"var comm ${serviceName}Comm.Handler" },
+    )
+
+  /** Given a service name, whether the service uses inter-service communication, the endpoints desired and the port
+    * number, generate the main function */
+  private[go] def generateMain(serviceName: String, usesComms: Boolean, endpoints: Set[Endpoint], port: Int): String = {
+    val sb = new StringBuilder
+    sb.append("func main() ")
```
c-harding: This is unused, right?

c-harding (commented):
Nice, just one redundant variable

src/main/scala/temple/generate/service/go/GoServiceMainGenerator.scala
```diff
@@ -0,0 +1,102 @@
...
+
```
c-harding:
```suggestion
```

Lewiky (approved):
OOOH YEAH. I LIKE THAT<><><<><>>>

src/main/scala/temple/generate/service/go/GoServiceMainGenerator.scala
```diff
@@ -0,0 +1,102 @@
...
+      "valid.SetFieldsRequiredByDefault(true)",
+      "",
+      "config, err := util.GetConfig(*configPtr)",
+      "if err != nil " + CodeWrap.curly.tabbed("log.Fatal(err)"),
+      "",
+      s"dao = ${serviceName}DAO.DAO{}",
+      "err = dao.Init(config)",
+      "if err != nil " + CodeWrap.curly.tabbed("log.Fatal(err)"),
+      "",
+      when[CodeTerm](usesComms) {
```
c-harding:
```suggestion
      when(usesComms) {
```

src/main/scala/temple/generate/service/go/GoServiceMainGenerator.scala
willjvsmith: Yeah this should have been removed in your refactor PR, I'll remove it here.

Though in general I've tried to keep the order of code generating the same as the order of code, i.e. `"fun main() "` and then the body

willjvsmith: I guess we don't need the `body` variable, and we can maintain the order

src/main/scala/temple/generate/service/go/GoServiceMainGenerator.scala
```diff
@@ -0,0 +1,102 @@
...
+        mkCode.lines(
+          s"comm = ${serviceName}Comm.Handler{}",
+          "comm.Init(config)",
+          "",
+        )
+      },
+      "r := mux.NewRouter()",
+      when[CodeTerm](endpoints.contains(Endpoint.ReadAll)) {
```
c-harding:
```suggestion
      when(endpoints.contains(Endpoint.ReadAll)) {
```

c-harding (commented):
The type parameters also aren’t needed here (as the contents are strings, and so don’t need to be cast)

src/main/scala/temple/generate/service/go/GoServiceMainGenerator.scala
c-harding: Yeah, this was my fault, sorry

c-harding (approved):

PR #111, opened 2020-03-06
Openapi gen

c-harding:
Surprised this is only 152 lines, was worried it was going to be bigger.

- Creates a function to generate the OpenAPI description for an error response type.
- Adds a library for Yaml-ifying it, as used in a test.
- Adds a trait for Jsonable structures with a custom Json representation
- I got very confused by the implicit nature of the `circe` library, but I think most of weirdest magic is factored out now. Now all that is needed is `import io.circe.syntax._` to get `.asJson` and `import io.circe.yaml.syntax.AsYaml` to get `.asYaml`. However, `temple.generate.target.openapi.Jsonable#encodeJsonable`/`temple.generate.target.openapi.Jsonable.encodeJsonable` is duplicated as protected to save on needing to import it elsewhere inside concrete `Jsonable` implementations.

Side tidying:

- Fixes a warning of some redundant parens in `build.sbt`

### DIFF
--- a/src/main/scala/temple/generate/target/openapi/Jsonable.scala
+++ b/src/main/scala/temple/generate/target/openapi/Jsonable.scala
@@ -1,0 +1,21 @@
+package temple.generate.target.openapi
+
+import io.circe.{Encoder, Json}
+
+private[openapi] trait Jsonable {
+
+  /** Turn a case class into a map in preparation for conversion to a JSON object */
+  def toJsonMap: Map[String, Option[Json]]
+
+  // Required so that nested Jsonable interfaces always call the correct nested version
+  implicit final protected def encodeJsonable[T <: Jsonable]: Encoder[T] = Jsonable.encodeJsonable
+}
+
+private[openapi] object Jsonable {
+
+  /** Create an encoder for JSON objects by providing a function to map them to options of values */
+  private def mapSequenceEncoder[T](f: T => Map[String, Option[Json]]): Encoder[T] =
+    (a: T) => Json.obj(f(a).iterator.flatMap { case (str, maybeJson) => maybeJson.map(str -> _) }.toSeq: _*)
+
+  implicit def encodeJsonable[T <: Jsonable]: Encoder[T] = mapSequenceEncoder(_.toJsonMap)
+}
--- a/src/main/scala/temple/generate/target/openapi/OpenAPIGenerator.scala
+++ b/src/main/scala/temple/generate/target/openapi/OpenAPIGenerator.scala
@@ -8,7 +8,7 @@ import scala.collection.immutable.ListMap
 object OpenAPIGenerator {
 
   /** Create a Response representation for an error */
-  private[openapi] def generateError(description: String, code: Int, example: String): Response =
+  private[openapi] def generateError(description: String, example: String): Response =
     Response(
       description,
       ListMap(
--- a/src/main/scala/temple/generate/target/openapi/OpenAPIGenerator.scala
+++ b/src/main/scala/temple/generate/target/openapi/OpenAPIGenerator.scala
@@ -1,0 +1,21 @@
+package temple.generate.target.openapi
+
+import temple.generate.target.openapi.OpenAPIType._
+import io.circe.syntax._
+
+import scala.collection.immutable.ListMap
+
+object OpenAPIGenerator {
+
+  /** Create a Response representation for an error */
+  private[openapi] def generateError(description: String, code: Int, example: String): Response =
+    Response(
+      description,
+      ListMap(
+        "application/json" -> Response.MediaTypeObject(
+          OpenAPIObject(ListMap("error" -> OpenAPISimpleType("string", "example" -> example.asJson))),
+        ),
+      ),
+    )
+
+}
--- a/src/main/scala/temple/generate/target/openapi/OpenAPIType.scala
+++ b/src/main/scala/temple/generate/target/openapi/OpenAPIType.scala
@@ -5,8 +5,8 @@ import io.circe.syntax._
 
 import scala.collection.immutable.ListMap
 
-sealed abstract private[openapi] class OpenAPIType(customFields: Seq[(String, Json)]) extends JsonEncodable {
-  def typeString: String
+sealed abstract private[openapi] class OpenAPIType(val typeString: String, customFields: Seq[(String, Json)])
+    extends JsonEncodable {
 
   override def toJsonMap: Map[String, Option[Json]] =
     ListMap("type" -> Some(typeString.asJson)) ++ customFields.to(ListMap).view.mapValues(Some(_))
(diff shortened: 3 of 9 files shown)


src/main/scala/temple/generate/target/openapi/Jsonable.scala
```diff
@@ -0,0 +1,21 @@
+package temple.generate.target.openapi
+
+import io.circe.{Encoder, Json}
+
+private[openapi] trait Jsonable {
```
jaylees14: How about `JsonEncodable`?

src/main/scala/temple/generate/target/openapi/Jsonable.scala
```diff
@@ -0,0 +1,21 @@
...
+
+  // Required so that nested Jsonable interfaces always call the correct nested version
+  implicit final protected def encodeJsonable[T <: Jsonable]: Encoder[T] = Jsonable.encodeJsonable
+}
+
+private[openapi] object Jsonable {
+
+  /** Create an encoder for JSON objects by providing a function to map them to options of values */
+  private def mapSequenceEncoder[T](f: T => Map[String, Option[Json]]): Encoder[T] =
+    (a: T) => Json.obj(f(a).iterator.flatMap { case (str, maybeJson) => maybeJson.map(str -> _) }.toSeq: _*)
```
jaylees14: Can we break this line up with some variable names? It's pretty hard to follow

src/main/scala/temple/generate/target/openapi/OpenAPIType.scala
```diff
@@ -0,0 +1,33 @@
...
+import scala.collection.immutable.ListMap
+
+sealed abstract private[openapi] class OpenAPIType(fieldEntries: Seq[(String, Json)]) extends Jsonable {
+  def typeString: String
+  final lazy val customFields: Map[String, Json] = fieldEntries.to(ListMap)
+
+  override def toJsonMap: Map[String, Option[Json]] =
+    ListMap(
+      "type"       -> Some(typeString.asJson),
+      "properties" -> this.matchPartial { case openAPIType: OpenAPIObject => openAPIType.properties.asJson },
```
jaylees14: Just a thought - is this good design, for the abstract "parent" type to know about the "children" types?

src/main/scala/temple/generate/target/openapi/Response.scala
```diff
@@ -0,0 +1,34 @@
...
+
+import io.circe.{Encoder, Json}
+
+import scala.collection.immutable.ListMap
+
+/** https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#responseObject */
+private[openapi] case class Response(
+  description: String,
+  content: Map[String, Response.MediaTypeObject],
+  required: Option[Boolean] = None,
```
jaylees14, line 13: Kinda annoying that you'd have to wrap it in `Some` to use it, but I see why

jaylees14 (commented):
Looks good, just a few general comments. Happy to approve if you don't agree tho 👍

src/main/scala/temple/generate/target/openapi/OpenAPIGenerator.scala
```diff
@@ -0,0 +1,21 @@
...
+
+import scala.collection.immutable.ListMap
+
+object OpenAPIGenerator {
+
+  /** Create a Response representation for an error */
+  private[openapi] def generateError(description: String, code: Int, example: String): Response =
+    Response(
+      description,
+      ListMap(
```
Lewiky, line 14: Did we work out that ListMap maintains dict order?

src/main/scala/temple/generate/target/openapi/OpenAPIType.scala
```diff
@@ -0,0 +1,33 @@
...
+      "items"      -> this.matchPartial { case openAPIType: OpenAPIArray => openAPIType.items.asJson },
```
Lewiky: matchPartial?

src/main/scala/temple/generate/target/openapi/OpenAPIType.scala
```diff
@@ -0,0 +1,33 @@
...
+    ) ++ customFields.view.mapValues(Some(_))
+}
+
+private[openapi] object OpenAPIType {
+
+  sealed case class OpenAPISimpleType(typeString: String, fieldEntries: (String, Json)*)
+      extends OpenAPIType(fieldEntries)
```
Lewiky: What is `typeString` in this case where it isn't defined? How does the `toJsonMap` method work in this case?

src/main/scala/temple/utils/MonadUtils.scala
```diff
@@ -8,4 +8,11 @@ object MonadUtils {
     /** Reduces the Either to an item of the right type, by transforming a Left item if present */
     def fromEither(f: A => B): B = either.fold(f, identity)
   }
+
+  /** Adds the method `matchPartial` to any object, to call a partial function directly on it */
+  implicit class MatchPartial[A](x: A) {
```
Lewiky: Ok so this is how it's defined, resolves my earlier comment but could we get a comment somewhere to point to MonadUtils?

src/main/scala/temple/generate/target/openapi/OpenAPIGenerator.scala
```diff
@@ -0,0 +1,21 @@
...
+
+import temple.generate.target.openapi.OpenAPIType._
+import io.circe.syntax._
+
+import scala.collection.immutable.ListMap
+
+object OpenAPIGenerator {
+
+  /** Create a Response representation for an error */
+  private[openapi] def generateError(description: String, code: Int, example: String): Response =
```
Lewiky: Where does `code` get used?

src/main/scala/temple/generate/target/openapi/OpenAPIGenerator.scala
c-harding: Yes

src/main/scala/temple/generate/target/openapi/OpenAPIType.scala
c-harding: How does this alternative look? Removes the need for matchPartial, as each child passes up the additional fields they define
https://github.com/TempleEight/temple/compare/openapi-gen...openapi-gen-rewrite-openapitype-tojsonmap-2

Another alternative, but personally I think this seems a little more repetitive:
https://github.com/TempleEight/temple/compare/openapi-gen...openapi-gen-rewrite-openapitype-tojsonmap

src/main/scala/temple/generate/target/openapi/OpenAPIType.scala
c-harding: Anything like number or string, e.g. in

https://github.com/TempleEight/temple/blob/5c637f8d7ed6ae7d72d0acfe30891632e924958f/src/main/scala/temple/generate/target/openapi/OpenAPIGenerator.scala#L16

src/main/scala/temple/generate/target/openapi/Response.scala
c-harding: Can you think of an alternative? Guess I could default to `false`, and only render it if `true`, but you may wish to be more explicit in some cases

src/main/scala/temple/generate/target/openapi/OpenAPIType.scala
c-harding: In terms of how it works, it’s been updated slightly, but originally the definition of a value in a case class overrides a def in the parent class

jaylees14 (approved):

Lewiky (approved):

PR #112, opened 2020-03-10
[openapi-gen] Refactor JsonEncodable

c-harding:
This provides versions with and without optionals for each value in the tree, and moves it up to the root `generation` package for Lewis

Code cov is reduced merely because the number of lines is reduced, and so the other untested lines are proportionally more significant


Lewiky (approved):
Yeah this is gooooood

Lewiky (approved):

jaylees14 (approved):

PR #115, opened 2020-03-11
[api] Add FlagMapView, for use with OpenAPI generation

c-harding:
```scala
  val errorTracker = FlagMapView(
    400 -> OpenAPIBuilder.generateError("Invalid request", "Invalid request parameters: name"),
    404 -> OpenAPIBuilder.generateError("ID not found", "Object not found with ID 1"),
    500 -> OpenAPIBuilder.generateError(
      "The server encountered an error while serving this request",
      "Unable to reach user service: connection timeout",
    ),
  )

  // while generating one of the endpoints:
  errorTracker.flag(500)
  val errorBlock = errorTracker.toMap.asJson // -> map just including the 500 key
```

Provide a maximal map at the start, and then specify which keys to access. This will be used for error references in the OpenAPI spec: the definition of all error codes will be provided, then they will be activated as and when the references are needed. This way, there won’t be any unused references (e.g. if no error 403 occur in the system, the 403 block won’t appear in the references block).

Other applications could include imports and initialization/closing code

### DIFF
--- a/src/main/scala/temple/collection/FlagMapView.scala
+++ b/src/main/scala/temple/collection/FlagMapView.scala
@@ -28,7 +28,7 @@ class FlagMapView[K, +V] private (val basis: Map[K, V], keyDesc: String = "flag"
   def flag(keys: K*): FlagMapView[K, V] = {
     keys.foreach { key =>
       if (!basis.contains(key))
-        throw new NoSuchElementException(s"The $keyDesc $key has not been defined")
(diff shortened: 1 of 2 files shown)


jaylees14 (approved):
Sure, code looks good, but feels like quite a lot of work to avoid some repeated lookups.

Also love the completely unreadable syntax for covariance 🙄

Lewiky (approved):

PR #116, opened 2020-03-11
[openapi-gen] Add data structures for full OpenAPI generation

c-harding:
Sorry for coverage drop, usage of these data structures coming very soon

### DIFF
--- a/src/main/scala/temple/generate/target/openapi/HTTPVerb.scala
+++ b/src/main/scala/temple/generate/target/openapi/HTTPVerb.scala
@@ -2,13 +2,14 @@ package temple.generate.target.openapi
 
 import io.circe.KeyEncoder
 
-object HTTPVerb extends Enumeration {
-  type HTTPVerb = Value
-  val Get: HTTPVerb    = Value("get")
-  val Put: HTTPVerb    = Value("put")
-  val Post: HTTPVerb   = Value("post")
-  val Patch: HTTPVerb  = Value("patch")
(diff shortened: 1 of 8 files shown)


src/main/scala/temple/generate/target/openapi/HTTPVerb.scala
```diff
@@ -0,0 +1,14 @@
...
+
+object HTTPVerb extends Enumeration {
+  type HTTPVerb = Value
+  val Get: HTTPVerb    = Value("get")
+  val Put: HTTPVerb    = Value("put")
+  val Post: HTTPVerb   = Value("post")
+  val Patch: HTTPVerb  = Value("patch")
+  val Delete: HTTPVerb = Value("delete")
+
+  implicit def httpVerbKeyEncoder: KeyEncoder[HTTPVerb] = _.toString
```
c-harding:
```suggestion
sealed abstract class HTTPVerb(val verb: String)

object HTTPVerb {
  case object Get    extends HTTPVerb("get")
  case object Put    extends HTTPVerb("put")
  case object Post   extends HTTPVerb("post")
  case object Patch  extends HTTPVerb("patch")
  case object Delete extends HTTPVerb("delete")

  implicit def httpVerbKeyEncoder: KeyEncoder[HTTPVerb] = _.verb
```

Lewiky (approved):

jaylees14 (approved):
Awesome, nice and clean

PR #117, opened 2020-03-12
[openAPI-gen] Generate Create endpoints in OpenAPI format

c-harding:
Same as #119 but for the create endpoint


src/main/scala/temple/generate/target/openapi/OpenAPIBuilder.scala
```diff
@@ -0,0 +1,148 @@
...
+      val maxLength = max.map("maxLength" -> _.asJson)
+      OpenAPISimpleType("string", Seq(minLength, maxLength).flatten: _*)
+    case IntType(max, min, precision) =>
+      val minimum = min.map("minimum" -> _.asJson)
+      val maximum = max.map("maximum" -> _.asJson)
+      OpenAPISimpleType("number", if (precision > 4) "int64" else "int32", Seq(minimum, maximum).flatten: _*)
+    case FloatType(max, min, precision) =>
+      val minimum = min.map("minimum" -> _.asJson)
+      val maximum = max.map("maximum" -> _.asJson)
+      OpenAPISimpleType("number", if (precision > 4) "double" else "float", Seq(minimum, maximum).flatten: _*)
```
c-harding: A little repetitive, but not sure how to avoid it. `Seq().flatten: _*` is the simplest way I could find to optionally supply arguments to a variadic function

src/main/scala/temple/generate/target/openapi/OpenAPIBuilder.scala
```diff
@@ -0,0 +1,148 @@
...
+    case ForeignKey(references) =>
+      OpenAPISimpleType("number", "int32", "description" -> s"Reference to $references ID".asJson)
+  }
+
+  private def generateItemType(attributes: Map[String, Attribute]): OpenAPIObject = OpenAPIObject(
+    attributes.iterator
+      .filter { case _ -> attribute => !isServerAttribute(attribute) }
+      .map { case str -> attribute => str -> attributeToOpenAPIType(attribute) }
+      .to(attributes.mapFactory),
```
c-harding: This produces a map of the same subtype as `attributes` started at

src/main/scala/temple/generate/target/openapi/OpenAPIBuilder.scala
```diff
@@ -0,0 +1,149 @@
...
+                jsonContent(MediaTypeObject(generateItemType(service.attributes))),
+                s"$capitalizedName successfully created",
+              ),
+              400 -> Response.Ref(useError(400)),
+              500 -> Response.Ref(useError(500)),
+            ),
+          )
+      case Read   => // TODO in future PR
+      case Update => // TODO in future PR
+      case Delete => // TODO in future PR
```
c-harding: Would each of these be better served in their own functions? Does mean I can’t prepare `lowerName`, `capitalizedName` etc in a similar place for all of them

jaylees14 (approved):
LGTM 💯

src/main/scala/temple/generate/target/openapi/OpenAPIBuilder.scala
jaylees14: Maybe, then they just take it as a parameter? Happy either way though

Lewiky (approved):

PR #118, opened 2020-03-12
[OpenAPI-gen] Finalise datastructures in preparation for generation

c-harding:
The PR for OpenAPI generation (#117) was looking too monstrous, so this should ease the pain slightly.

- `Info` now has a custom renderer, so description and version are omitted if empty
- `OpenAPISimpleType` has an optional argument `format` as it was being used so often
- A `Service` datatype was created, which will form the input of the OpenAPI generator


jaylees14 (approved):
Beautiful

willjvsmith (approved):

PR #119, opened 2020-03-12
[openAPI-gen] Generate the first endpoint in OpenAPI format

c-harding:
This is a slightly cut-down version of #117, and will be even smaller once #118 is merged.

There are +118 lines of tests, +34–11 lines as part of #118, and +129–39 lines of main code in this PR.

`OpenAPIBuilder.generateError` and `OpenAPIBuilder should generate error descriptions correctly` are carried over from deleted files, but because of the small similarity Git can’t detect the continuity.

Main feature: `OpenAPIBuilder.render`, for turning a description and a list of services into an OpenAPI specification file, at present only rendering the ReadAll endpoint.


willjvsmith (approved):
LGTM 🚀

src/main/scala/temple/generate/target/openapi/OpenAPIBuilder.scala
```diff
@@ -0,0 +1,129 @@
...
+import temple.DSL.semantics.{Annotation, Attribute}
+import temple.collection.FlagMapView
+import temple.generate.Endpoint._
+import temple.generate.target.openapi.OpenAPIFile.{Components, Info}
+import temple.generate.target.openapi.OpenAPIType._
+
+import scala.collection.immutable.ListMap
+import scala.collection.mutable
+
+private class OpenAPIBuilder private (name: String, version: String, description: String = "") {
```
jaylees14: I'm not sure `Builder` is the right term - we've been using `Generator` everywhere else (e.g. `PostgresGenerator`, `DockerfileGenerator`), and leaving `Builder` for when we do the mapping?

src/main/scala/temple/generate/target/openapi/OpenAPIBuilder.scala
```diff
@@ -0,0 +1,129 @@
...
+
+  private def jsonContent(mediaTypeObject: MediaTypeObject) = Map("application/json" -> mediaTypeObject)
+
+  def addPaths(service: Service): this.type = {
+    val lowerName       = service.name.toLowerCase
+    val capitalizedName = service.name.capitalize
+    val tags            = Seq(capitalizedName)
+    service.endpoints.foreach {
+      case ReadAll =>
+        path(s"/$lowerName/all") += HTTPVerb.Get -> Path(
```
jaylees14: This reads a little confusingly, you're calling a function and then adding to it? Took me a while to realise that this will implicitly update the `paths` variable

src/main/scala/temple/generate/target/openapi/OpenAPIBuilder.scala
```diff
@@ -0,0 +1,129 @@
...
+    services.foreach(builder.addPaths)
+    builder.toOpenAPI
+  }
+
+  def render(name: String, version: String, description: String = "")(services: Service*): String =
+    Printer(preserveOrder = true, dropNullKeys = true).pretty(build(name, version, description)(services: _*).asJson)
+
+  /** Create a Response representation for an error */
+  private[openapi] def generateError(description: String, example: String): Response =
+    Literal(
```
jaylees14: `Literal` isn't very descriptive, wonder if there's a better name for this?

src/main/scala/temple/generate/target/openapi/OpenAPIBuilder.scala
```diff
@@ -0,0 +1,129 @@
...
+      description = description,
+      content = ListMap(
+        "application/json" -> MediaTypeObject(
```
jaylees14: Do you not want to use your `jsonContent` function here? Seems silly to have it but then not use it because it's private

jaylees14 (commented):
Couple of thoughts, looks good though

jaylees14 (approved):
Awesome stuff

PR #120, opened 2020-03-13
[OpenAPI-gen] Add parent trait for JsonEncodable, so non-objects can be encoded

c-harding:
Small change to JsonEncodable, so that JsonEncodable can be used for non-objects (strings, arrays etc). This is so we can encode case objects as strings by extending `JsonEncodable` directly, defining `toJson`.


Lewiky (approved):
LGTM

jaylees14 (approved):

PR #121, opened 2020-03-13
Json encodable arrows

c-harding:
Depends on #120, but reduces all the needless .asJson in `jsonEntryIterator` and `jsonOptionEntryIterator`, by providing a custom arrow `~>`/`~~>` that inserts `.asJson`/`.map(_.asJson)`. Note that because asJson takes an implicit argument of the encoder, the arrow itself must be replaced instead, instead of dealing with the subsequently produced sequence of heterogeneously typed tuples.

### DIFF
--- a/src/main/scala/temple/generate/JsonEncodable.scala
+++ b/src/main/scala/temple/generate/JsonEncodable.scala
@@ -1,5 +1,6 @@
 package temple.generate
 
+import io.circe.syntax._
 import io.circe.{Encoder, Json}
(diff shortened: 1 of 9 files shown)


Lewiky (approved):
Love this

jaylees14 (approved):
Yeah, I'm dubious about adding sugar everywhere because it increases the barrier to entry for a developer who wants to understand and modify this code, but the changes look good.

PR #122, opened 2020-03-13
[K8s-Gen] Generate Database Deployment Scripts

Lewiky:
This refactors the existing Deployment generation code to also allow for database deployments to be built, which should also mean generalisation to more complicated deployments in the future should be achievable. 

Also implements the DB deployment generation as per the sample kube service

### DIFF
--- a/src/main/scala/temple/generate/kube/KubernetesGenerator.scala
+++ b/src/main/scala/temple/generate/kube/KubernetesGenerator.scala
@@ -2,6 +2,7 @@ package temple.generate.kube
 
 import io.circe.generic.auto._
 import io.circe.syntax._
+import io.circe.yaml.Printer
 import io.circe.yaml.syntax.AsYaml
 import temple.generate.FileSystem._
 import temple.generate.kube.ast.OrchestrationType._
@@ -26,7 +27,7 @@ object KubernetesGenerator {
     }
     val name = service.name + { if (isDb) "-db" else "" }
 
-    Header(version, kind, Metadata(name, Labels(name, genType, isDb))).asJson.asYaml.spaces2
+    Header(version, kind, Metadata(name, Labels(service.name, genType, isDb))).asJson.asYaml.spaces2
   }
 
   private def generateDbStorage(service: Service): String =
@@ -39,8 +40,46 @@ object KubernetesGenerator {
   private def generateDbService(service: Service): String =
     generateHeader(service, GenType.Service, isDb = true)
 
-  private def generateDbDeployment(service: Service): String =
-    generateHeader(service, GenType.Deployment, isDb = true)
+  private def generateDbDeployment(service: Service): String = {
+    val name = service.name + "-db"
+    val deploymentBody = Body(
+      DeploymentSpec(
+        1,
+        Selector(Labels(service.name, GenType.Deployment, isDb = true)),
+        strategy = Some(Strategy("Recreate")),
+        Template(
+          Metadata(name, Labels(service.name, GenType.Deployment, isDb = true)),
+          PodSpec(
+            name,
+            Seq(
+              Container(
+                service.dbImage,
+                name,
+                ports = Seq(),
+                env = service.envVars.map(x => EnvVar(x._1, x._2)),
+                volumeMounts = Seq(
+                  VolumeMount("/var/lib/postgresql/data", None, name + "-claim"),
+                  VolumeMount("/docker-entrypoint-initdb.d/init.sql", Some("init.sql"), name + "-init"),
+                ),
+                lifecycle = Some(Lifecycle("echo done")),
+              ),
+            ),
+            imagePullSecrets = Seq(),
+            restartPolicy = "Always",
+            volumes = Seq(
+              Volume(name + "-init", ConfigMap(name + "-config")),
+              Volume(name + "-claim", PersistentVolume(name + "-claim")),
+            ),
+          ),
+        ),
+      ),
+    ).asJson
+    val printer = Printer(preserveOrder = true, dropNullKeys = true)
+    mkCode(
+      generateHeader(service, GenType.Deployment, isDb = true),
+      printer.pretty(deploymentBody),
+    )
+  }
 
   private def generateService(service: Service): String = {
     val serviceBody = Body(
@@ -60,13 +99,23 @@ object KubernetesGenerator {
       DeploymentSpec(
         service.replicas,
         Selector(Labels(service.name, GenType.Deployment, isDb = false)),
+        strategy = None,
         Template(
           Metadata(service.name, Labels(service.name, GenType.Deployment, isDb = false)),
           PodSpec(
             service.name,
-            Seq(Container(service.image, service.name, service.ports.map(x => ContainerPort(x._2)))),
+            Seq(
+              Container(
+                service.image,
+                service.name,
+                service.ports.map(x => ContainerPort(x._2)),
+                env = Seq(),
+                volumeMounts = Seq(),
+              ),
+            ),
             Seq(Secret(service.secretName)),
             restartPolicy = "Always",
+            volumes = Seq(),
           ),
         ),
       ),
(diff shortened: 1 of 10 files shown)


src/main/scala/temple/generate/kube/KubernetesGenerator.scala
```diff
@@ -26,7 +27,7 @@ object KubernetesGenerator {
     }
     val name = service.name + { if (isDb) "-db" else "" }
```
c-harding, line 31: is this used?

c-harding (approved):
lgtm, just the unused variable

src/main/scala/temple/generate/kube/KubernetesGenerator.scala
```diff
@@ -39,8 +40,46 @@ object KubernetesGenerator {
...
-    generateHeader(service, GenType.Deployment, isDb = true)
+  private def generateDbDeployment(service: Service): String = {
+    val name = service.name + "-db"
+    val deploymentBody = Body(
+      DeploymentSpec(
+        1,
+        Selector(Labels(service.name, GenType.Deployment, isDb = true)),
+        strategy = Some(Strategy("Recreate")),
+        Template(
+          Metadata(name, Labels(service.name, GenType.Deployment, isDb = true)),
```
jaylees14: I'm wondering if breaking this up into some variables might be nice?

src/main/scala/temple/generate/kube/KubernetesGenerator.scala
```diff
@@ -39,8 +40,46 @@ object KubernetesGenerator {
...
+            restartPolicy = "Always",
+            volumes = Seq(
+              Volume(name + "-init", ConfigMap(name + "-config")),
+              Volume(name + "-claim", PersistentVolume(name + "-claim")),
+            ),
+          ),
+        ),
+      ),
+    ).asJson
+    val printer = Printer(preserveOrder = true, dropNullKeys = true)
```
jaylees14: Why are we using a `printer` here but not elsewhere?

src/main/scala/temple/generate/kube/ast/OrchestrationType.scala
```diff
@@ -9,10 +9,20 @@ object OrchestrationType {
...
-    * @param secretName The name of the Kubernetes secret used to fetch images from the registr
+    * @param secretName The name of the Kubernetes secret used to fetch images from the registry
+    * @param envVars A sequence of key -> value to set as environment variables in the container
     */
-  case class Service(name: String, image: String, ports: Seq[(String, Int)], replicas: Int, secretName: String)
+  case class Service(
+    name: String,
+    image: String,
+    dbImage: String,
+    ports: Seq[(String, Int)],
```
jaylees14, line 26: Is there an advantage to using a `Seq` over a `Map`?

src/main/scala/temple/generate/kube/KubernetesGenerator.scala
```diff
@@ -39,8 +40,46 @@ object KubernetesGenerator {
...
+            name,
+            Seq(
+              Container(
+                service.dbImage,
+                name,
+                ports = Seq(),
+                env = service.envVars.map(x => EnvVar(x._1, x._2)),
+                volumeMounts = Seq(
+                  VolumeMount("/var/lib/postgresql/data", None, name + "-claim"),
+                  VolumeMount("/docker-entrypoint-initdb.d/init.sql", Some("init.sql"), name + "-init"),
```
jaylees14: These are making assumptions about the selected database: you should be taking that as an argument and switching on it

src/main/scala/temple/generate/kube/ast/gen/Spec.scala
```diff
@@ -4,33 +4,148 @@ import io.circe.Json
...
 
-  case class Selector(matchLabels: Labels)
+    override def jsonOptionEntryIterator: IterableOnce[(String, Option[Json])] = Seq(
+      "ports" ~~> when(ports.nonEmpty)(ports),
+      "selector" ~~> Some(selector),
+    )
+  }
+
+  case class Selector(matchLabels: Labels) extends JsonEncodable.Object {
+    override def jsonEntryIterator: IterableOnce[(String, Json)] = Seq("matchLabels" ~> matchLabels)
```
jaylees14, line 36: Think you need whitespace above this line

src/main/scala/temple/generate/kube/KubernetesGenerator.scala
```diff
@@ -60,13 +99,23 @@ object KubernetesGenerator {
...
         Template(
           Metadata(service.name, Labels(service.name, GenType.Deployment, isDb = false)),
           PodSpec(
             service.name,
-            Seq(Container(service.image, service.name, service.ports.map(x => ContainerPort(x._2)))),
+            Seq(
+              Container(
+                service.image,
+                service.name,
+                service.ports.map(x => ContainerPort(x._2)),
```
jaylees14: Not a fan of the`._2`, doesn't really mean much. How about we have a type with some named arguments? @xsanda

src/main/scala/temple/generate/kube/KubernetesGenerator.scala
```diff
@@ -39,8 +40,46 @@ object KubernetesGenerator {
...
 
-  private def generateDbDeployment(service: Service): String =
-    generateHeader(service, GenType.Deployment, isDb = true)
+  private def generateDbDeployment(service: Service): String = {
+    val name = service.name + "-db"
+    val deploymentBody = Body(
+      DeploymentSpec(
+        1,
+        Selector(Labels(service.name, GenType.Deployment, isDb = true)),
+        strategy = Some(Strategy("Recreate")),
```
jaylees14: Wondering if `Strategy` could be predefined? And we just use `Strategy.Recreate`?

src/main/scala/temple/generate/kube/KubernetesGenerator.scala
```diff
@@ -39,8 +40,46 @@ object KubernetesGenerator {
...
+                env = service.envVars.map(x => EnvVar(x._1, x._2)),
+                volumeMounts = Seq(
+                  VolumeMount("/var/lib/postgresql/data", None, name + "-claim"),
+                  VolumeMount("/docker-entrypoint-initdb.d/init.sql", Some("init.sql"), name + "-init"),
+                ),
+                lifecycle = Some(Lifecycle("echo done")),
+              ),
+            ),
+            imagePullSecrets = Seq(),
+            restartPolicy = "Always",
```
jaylees14: Also here, `RestartPolicy.Always`? Avoids typos etc.

src/test/resources/kube/user-db-deployment.yaml
```diff
@@ -0,0 +1,50 @@
...
+        - mountPath: /docker-entrypoint-initdb.d/init.sql
+          subPath: init.sql
+          name: user-db-init
+        lifecycle:
+          postStart:
+            exec:
+              command:
+              - /bin/sh
+              - -c
+              - echo done
```
jaylees14, line 42: We currently use `["/bin/sh", "-c", "echo done"]`, does this work the same?

jaylees14 (commented):
Couple of comments, but looks good

src/main/scala/temple/generate/kube/KubernetesGenerator.scala
Lewiky: Yeah it's used in the line below

src/main/scala/temple/generate/kube/ast/OrchestrationType.scala
Lewiky: We only ever need to iterate/map over the ports so a list of tuples is really all we need, however I'm happy to change to using Map[K,V] everywhere instead of Seq[(K, V)] If we think it's more readable

jaylees14: That's cool, was more curiosity than anything else

src/test/resources/kube/user-db-deployment.yaml
Lewiky: Yeah it makes no difference, tested to verify

jaylees14 (approved):
Nice one 👍

PR #124, opened 2020-03-16
[K8s-Gen] DB-Service and Storage Generation

Lewiky:
Final 'dynamic' kube-gen PR. This adds support for database services, and the storage config for DB pods.


jaylees14 (approved):
Had to do review IRL, but aside from the `capacity` rename, LGTM

src/main/scala/temple/generate/kube/KubernetesGenerator.scala
```diff
@@ -28,20 +29,56 @@ object KubernetesGenerator {
       case GenType.StorageClaim => "PersistentVolumeClaim"
       case GenType.StorageMount => "PersistentVolume"
     }
-    val name = service.name + { if (isDb) "-db" else "" }
+    val name = service.name + (genType match {
+        case GenType.StorageClaim => "-db-claim"
+        case GenType.StorageMount => "-db-volume"
+        case _                    => if (isDb) "-db" else ""
```
c-harding: This line should likely be split into

```scala
case _ if isDb => "-db"
case _         => ""
```
```suggestion
        case if isDb              => "-db"
        case _                    => ""
```

It would also be nice if this could be stored in a variable (`suffix?`), `match` looks a bit odd in parens

src/main/scala/temple/generate/kube/KubernetesGenerator.scala
```diff
@@ -28,20 +29,56 @@ object KubernetesGenerator {
...
     this.printer.pretty(Header(version, kind, Metadata(name, Labels(service.name, genType, isDb))).asJson)
   }
 
-  private def generateDbStorage(service: Service): String =
-    mkCode.lines(
+  private def generateDbStorage(service: Service): String = {
+    val volumeBody = Body(
+      PersistentVolumeSpec(
+        storageClass = StorageClass.Manual,
+        capacityGb = 1.0f, //Gi,
```
c-harding:
```suggestion
        capacityGb = 1.0f, //GiB,
```

src/main/scala/temple/generate/kube/KubernetesGenerator.scala
```diff
@@ -85,7 +122,7 @@ object KubernetesGenerator {
 
     mkCode(
       generateHeader(service, GenType.Deployment, isDb = true),
-      printer.pretty(deploymentBody),
+      this.printer.pretty(deploymentBody),
```
c-harding, line 128: Is there a reason why we’re using `this` everywhere now? Is there another printer in scope locally?

willjvsmith (approved):

src/main/scala/temple/generate/kube/ast/gen/Spec.scala
```diff
@@ -31,6 +34,44 @@ object Spec {
     )
   }
 
+  case class PersistentVolumeSpec(
+    storageClass: StorageClass,
+    capacityGb: Float,
```
c-harding: Please can GB, MB etc be capitalised, to avoid confusion with gigabits?

src/main/scala/temple/generate/kube/ast/gen/volume/AccessMode.scala
```diff
@@ -0,0 +1,6 @@
+package temple.generate.kube.ast.gen.volume
+
+object AccessMode {
+  type AccessMode = String
+  val ReadWriteMany: AccessMode = "ReadWriteMany"
```
c-harding: These don’t feel particularly typesafe, perhaps a native Enumeration would work better? Native enums `toString` to their name

c-harding (commented):
A few small things, happy to accept if you disagree

src/main/scala/temple/generate/kube/KubernetesGenerator.scala
Lewiky: I just noticed I'd used both in different places and chose one for consistency, happy to change if needed

src/main/scala/temple/generate/kube/ast/OrchestrationType.scala
```diff
@@ -28,7 +26,7 @@ object OrchestrationType {
     secretName: String,
     envVars: Seq[(String, String)],
     dbStorage: DbStorage,
-    dbLifecycleCommand: LifecycleCommand,
+    dbLifecycleCommand: String,
```
c-harding, line 29: Is there a reason this isn’t `LifecycleCommand`?

c-harding (approved):

src/main/scala/temple/generate/kube/ast/OrchestrationType.scala
Lewiky: There was a weird bug I couldn't work out when it was, something about scala not being able to find the constructor to `Service` that matched the types, changing to a string worked - plus semantically that value _can_ be anything, so making it a string allows flexibility?

PR #126, opened 2020-03-16
[go-gen] Setup auth generation and generate auth imports

willjvsmith:
* Splits up Auth service generation from other service generation
* Moves generation common to both generating Auth and other services to `GoCommonGenerator`
* Adds import generation for Auth service
* Adds go.mod generation for Auth service
* Adds unit tests

Rather than having to shoehorn the Auth service generation into the other services and have to switch all over the place, I think it's quicker and less painful to separate it.


jaylees14 (approved):
LGTM 👍

Lewiky (approved):
Yeah boi LGTM

PR #129, opened 2020-03-16
[K8s-Gen] Kong Files

Lewiky:
Adds the static kong files into the generated stuff, just reads them from the resources and outputs them.

As such most of the bulk just comes from the kong yaml files.

Not much logic to test here, but getting close to being able to do some integration tests.


src/main/scala/temple/generate/kube/KubernetesGenerator.scala
```diff
@@ -181,6 +182,14 @@ object KubernetesGenerator {
     )
   }
 
+  private def buildKongFiles(): Map[File, FileContent] = Map(
```
c-harding: This shouldn’t have empty parens (as no side effects); it feels like it should be a `val` as it is just a static value

src/main/scala/temple/generate/kube/KubernetesGenerator.scala
```diff
@@ -181,6 +182,14 @@ object KubernetesGenerator {
...
+    File("kube/kong", "kong-db-deployment.yaml") -> FileUtils.readResources("kube/kong/kong-db-deployment.yaml"),
+    File("kube/kong", "kong-db-service.yaml")    -> FileUtils.readResources("kube/kong/kong-db-service.yaml"),
+    File("kube/kong", "kong-deployment.yaml")    -> FileUtils.readResources("kube/kong/kong-deployment.yaml"),
+    File("kube/kong", "kong-migration-job.yaml") -> FileUtils.readResources("kube/kong/kong-migration-job.yaml"),
+    File("kube/kong", "kong-service.yaml")       -> FileUtils.readResources("kube/kong/kong-service.yaml"),
+  )
+
   /** Given an [[OrchestrationRoot]], check the services inside it and generate deployment scripts */
   def generate(orchestrationRoot: OrchestrationRoot): Map[File, FileContent] =
     orchestrationRoot.services.flatMap { service =>
```
c-harding: Could this flatMap argument be extracted to a function?

c-harding (approved):
Nice, just some pondering about code layout

jaylees14 (approved):
Agree with @xsanda, but LGTM

PR #130, opened 2020-03-16
[go-gen] Generate handler.go for Auth service aside from Init function

willjvsmith:
* Moves `generateInit` for the `comm/handler.go`s to `GoCommonGenerator`
  * At the moment `GoCommonGenerator` is small, but if it gets bigger I'll split it up into `Main`, `Comm`, `DAO`, just as I have done for the other generation

The `init` function is not generated in this PR, as I want to use the common function, but `init` function has changed since the last release. Therefore changing it will break the old generation and their accompanying integration tests. Best do it in a separate PR.


src/main/resources/go/genFiles/auth/comm/create_jwt_credential.go
```diff
@@ -0,0 +1,70 @@
...
+	}
+
+	return &jwt, nil
+}
+
+// CreateJWTCredential provisions a new HS256 JWT credential
+func (comm *Handler) CreateJWTCredential() (*JWTCredential, error) {
+	hostname, ok := comm.Services["kong-admin"]
+	if !ok {
+		return nil, fmt.Errorf("service %s's hostname not in config file", "kong-admin")
```
jaylees14: Can we remove this unnecessary `%s`

src/main/scala/temple/generate/service/go/GoServiceGenerator.scala
```diff
@@ -55,7 +55,7 @@ object GoServiceGenerator extends ServiceGenerator {
         GoCommonGenerator.generatePackage("comm"),
         GoServiceCommGenerator.generateImports(serviceRoot.module),
         GoServiceCommGenerator.generateStructs(),
-        GoServiceCommGenerator.generateInit(),
+        GoCommonGenerator.generateCommInit(),
```
jaylees14, line 58: I'm struggling to see how a `CommInit` can be `Common`?

willjvsmith:
```
// Init sets up the Handler object with a list of services from the config
func Init(config *util.Config) *Handler {
	return &Handler{config.Services}
}
```

This is the same regardless of the service, including Auth, so can use the same function to read the same file.

jaylees14: I see, that was quite hard to follow from the naming

src/main/resources/go/genFiles/auth/comm/create_jwt_credential.go
willjvsmith: Needs to be removed spec-golang too, will make a card with an update for it

src/main/scala/temple/generate/service/go/GoServiceGenerator.scala
jaylees14: `GoCommon` sounds like it can be used anywhere, whereas I think what you want is `GoCommGenerator`, which then gets extended by `GoServiceCommGenerator` and `GoAuthCommGenerator`

jaylees14 (approved):

src/main/scala/temple/generate/service/go/GoServiceGenerator.scala
willjvsmith: See PR description, will address in next PR

Lewiky (approved):

PR #133, opened 2020-03-17
[OpenAPI-gen] Rename Endpoint to CRUD/Operation

c-harding:
This is to free up the name `Endpoint` to describe a concrete implementation at a URL


jaylees14 (approved):
I'd have personally gone with `CRUD`, since it's an acronym, but sure

willjvsmith (approved):
LGTM 🚀

PR #134, opened 2020-03-17
[go-gen] Finish refactor and clear main functions for integration tests

willjvsmith:
* Removes miscellaneous `println` in `SwaggerSpec`
* Renames all test go files to snippet, just the same as genFiles
* Makes all `comm/handler.go`s use the same `Init` generator function
  * This breaks tests, so cleared the main functions so that integration tests can still be used to validate the Go being generated
  * Also means temporarily taking out imports so validator doesn't complain about them being unused
* Reorders files in genFiles to exactly match generator hierarchy


c-harding (approved):
Nice and simple

Lewiky (approved):
cool

jaylees14 (approved):
LGTM 👍

PR #143, opened 2020-03-19
[K8s-Gen] Kube Map

Lewiky:
Adds the mapping between `Templefile` and `KubeGen`. - Most of the diff is adding all of the yaml test files for the `ProjectBuilder` Unit and E2E tests, can probably be skipped


src/e2e/scala/temple/SimpleE2ETest.scala
```diff
@@ -24,8 +23,8 @@ class SimpleE2ETest extends FlatSpec with Matchers {
       ),
     )
 
-    // Two folders should have been generated
-    val expectedFolders = Set("templeuser-db", "templeuser").map(dir => basePath.resolve(dir))
+    // Four folders should have been generated
```
c-harding:
```suggestion
    // Exactly these folders should have been generated
```

Prevents them getting out of sync, and clarifies that it’s not just the number of folders, we are checking which ones are generated

src/main/scala/temple/builder/KubernetesBuilder.scala
```diff
@@ -0,0 +1,38 @@
+package temple.builder
+
+import temple.DSL.semantics.Metadata.Database
+import temple.DSL.semantics.{Metadata, ServiceBlock}
```
c-harding:
```suggestion
import temple.ast.Metadata.Database
import temple.ast.{Metadata, ServiceBlock}
```
The merge 6e84646 broke this

src/e2e/scala/temple/SimpleE2ETest.scala
```diff
@@ -24,8 +23,8 @@ class SimpleE2ETest extends FlatSpec with Matchers {
...
+    val expectedFolders = Set("templeuser-db", "templeuser", "kong", "kube").map(dir => basePath.resolve(dir))
```
jaylees14, line 28: Should we not be checking the content of these folders?

src/main/scala/temple/builder/project/ProjectBuilder.scala
```diff
@@ -35,6 +37,16 @@ object ProjectBuilder {
         (File(s"${name.toLowerCase}", "Dockerfile"), dockerfileContents)
     }
 
-    Project(databaseCreationScripts ++ dockerfiles)
+    val kubeFiles = KubernetesGenerator.generate(
+      KubernetesBuilder
+        .createServiceKubeFiles(
+          templefile.projectName,
+          (templefile.services.zip(1024 until Int.MaxValue) map {
```
c-harding: Adding a dot before map means you don’t need the surrounding parens, may make it nicer

src/main/scala/temple/builder/project/ProjectBuilder.scala
```diff
@@ -35,6 +37,16 @@ object ProjectBuilder {
         (File(s"${name.toLowerCase}", "Dockerfile"), dockerfileContents)
     }
 
-    Project(databaseCreationScripts ++ dockerfiles)
+    val kubeFiles = KubernetesGenerator.generate(
+      KubernetesBuilder
```
jaylees14: Can we follow the pattern above and separate these into two variables?

src/main/scala/temple/builder/project/ProjectBuilder.scala
jaylees14: Also probably worth using the same one as above?

src/main/scala/temple/generate/kube/KubernetesGenerator.scala
```diff
@@ -91,7 +91,7 @@ object KubernetesGenerator {
       service.dbImage,
       name,
       ports = Seq(),
-      env = service.envVars.map(x => EnvVar(x._1, x._2)),
+      env = service.dbEnvVars.map(x => EnvVar(x._1, x._2)),
```
c-harding: Why doesn’t Service already contain EnvVars rather than tuples?

src/main/scala/temple/generate/kube/ast/OrchestrationType.scala
```diff
@@ -7,16 +7,18 @@ object OrchestrationType {
...
+    * @param dbImage            The name of the docker image to use for the service's datastore
+    * @param ports              The ports the Pod should expose, including the names of the ports (i.e www -> 80)
+    * @param replicas           The number of replicas of the pod that should be exposed
+    * @param secretName         The name of the Kubernetes secret used to fetch images from the registry
+    * @param appEnvVars         A sequence of key -> value to set as environment variables in the app container
+    * @param dbEnvVars          A sequence of key -> value to set as environment variables in the db container
+    * @param dbStorage          Object that defines how the service db should store data
     * @param dbLifecycleCommand The command to be ran in the database container on startup e.g to init the db
-    * @param usesAuth Whether this service requires authentication with the API Gateway before requests
+    * @param usesAuth           Whether this service requires authentication with the API Gateway before requests
```
c-harding, line 21: Is this spacing normal?

src/test/scala/temple/builder/project/ProjectBuilderTest.scala
```diff
@@ -9,33 +9,36 @@ class ProjectBuilderTest extends FlatSpec with Matchers {
...
     val project = ProjectBuilder.build(ProjectBuilderTestData.simpleTemplefile)
-    project.files shouldBe Map(
-      File("templeuser-db", "init.sql") -> ProjectBuilderTestData.simpleTemplefilePostgresCreateOutput,
-      File("templeuser", "Dockerfile")  -> ProjectBuilderTestData.simpleTemplefileUsersDockerfile,
-    )
+    val expected = Map(
+        File("templeuser-db", "init.sql") -> ProjectBuilderTestData.simpleTemplefilePostgresCreateOutput,
+        File("templeuser", "Dockerfile")  -> ProjectBuilderTestData.simpleTemplefileUsersDockerfile,
+      ) ++ ProjectBuilderTestData.simpleTemplefileKubeScripts
+    project.files.toSet shouldBe expected.toSet
```
c-harding: Does this work without the toSet?

c-harding (approved):
Few comments, looks nice

src/test/scala/temple/builder/project/ProjectBuilderTest.scala
```diff
@@ -9,33 +9,36 @@ class ProjectBuilderTest extends FlatSpec with Matchers {
...
   it should "correctly create a simple project using postgres as the default" in {
     val project = ProjectBuilder.build(ProjectBuilderTestData.simpleTemplefile)
-    project.files shouldBe Map(
-      File("templeuser-db", "init.sql") -> ProjectBuilderTestData.simpleTemplefilePostgresCreateOutput,
-      File("templeuser", "Dockerfile")  -> ProjectBuilderTestData.simpleTemplefileUsersDockerfile,
-    )
+    val expected = Map(
+        File("templeuser-db", "init.sql") -> ProjectBuilderTestData.simpleTemplefilePostgresCreateOutput,
+        File("templeuser", "Dockerfile")  -> ProjectBuilderTestData.simpleTemplefileUsersDockerfile,
+      ) ++ ProjectBuilderTestData.simpleTemplefileKubeScripts
```
jaylees14: Hmm not sure I'm a fan of this, would prefer to explicitly list every file - with its contents being the thing in `TestData`

src/test/scala/temple/builder/project/ProjectBuilderTestData.scala
```diff
@@ -93,6 +96,184 @@ object ProjectBuilderTestData {
...
+        |  labels:
+        |    app: TempleUser
+        |    kind: service
+        |spec:
+        |  replicas: 1
+        |  selector:
+        |    matchLabels:
+        |      app: TempleUser
+        |      kind: service
+        |  template:
```
jaylees14: Yeah this is quite hard to follow - can we break it into separate variables for each file's content

src/main/scala/temple/generate/kube/KubernetesGenerator.scala
Lewiky: The way I saw it - the `kube.Spec.EnvVar` here have a certain JSON encoding defined on them that's pretty specific to KubeGen, whereas the `OrchestrationType.Service` above should be generic to all orchestration

src/main/scala/temple/generate/kube/ast/OrchestrationType.scala
Lewiky: Think that's how scalafmt wants them

src/main/scala/temple/builder/project/ProjectBuilder.scala
```diff
@@ -35,6 +37,16 @@ object ProjectBuilder {
...
+        .createServiceKubeFiles(
```
c-harding: Please could this go on the same line as `KubernetesBuilder`, the chaining doesn’t make sense if there isn’t a flow of information (KubernetesBuilder isn’t some data to be passed down the pipeline)

src/main/scala/temple/builder/project/ProjectBuilder.scala
```diff
@@ -41,9 +40,12 @@ object ProjectBuilder {
...
         .createServiceKubeFiles(
           templefile.projectName,
-          (templefile.services.zip(1024 until Int.MaxValue) map {
-            case ((name, service), port) => (name, service, port)
-          }).toSeq,
+          templefile.services
+            .zip(1024 until Int.MaxValue)
+            .map {
+              case ((name, service), port) => (name, service, port)
+            }
```
c-harding: Could be one-lined

xsanda: In fact, having this defined on the Templefile might be useful
```scala
def servicesWithPorts: Iterable[(String, ServiceBlock, Int)] =
  templefile.services
    .zip(1024 until Int.MaxValue)
    .map { case ((name, service), port) => (name, service, port) }
```

src/main/scala/temple/builder/project/ProjectBuilder.scala
```diff
@@ -35,6 +36,19 @@ object ProjectBuilder {
...
-    Project(databaseCreationScripts ++ dockerfiles)
+    val kubeFiles = KubernetesGenerator.generate(
+      KubernetesBuilder
+        .createServiceKubeFiles(
+          templefile.projectName,
+          templefile.services
+            .zip(1024 until Int.MaxValue)
+            .map {
+              case ((name, service), port) => (name, service, port)
+            }
```
c-harding: In fact, having this defined on the Templefile might be useful, to avoid reuse of the magic number 1024
```scala
def servicesWithPorts: Iterable[(String, ServiceBlock, Int)] =
  templefile.services
    .zip(1024 until Int.MaxValue)
    .map { case ((name, service), port) => (name, service, port) }
```

src/main/scala/temple/generate/kube/ast/OrchestrationType.scala
c-harding: Mine doesn’t, but sure

src/main/scala/temple/generate/kube/KubernetesGenerator.scala
c-harding: Ok, fair enough. Worth noting that `service.dbEnvVars.map(EnvVar.tupled)` is equivalent, and is arguably nicer than listing each component with `_i`

src/main/scala/temple/builder/project/ProjectBuilder.scala
c-harding: Also wondering about `Iterator from 1024`/`Iterator.from(1024)` instead of explicitly stating the max value

src/main/scala/temple/builder/project/ProjectBuilder.scala
```diff
@@ -28,13 +29,23 @@ object ProjectBuilder {
         }
     }
 
-    val dockerfiles = templefile.services.zip(1024 until Int.MaxValue).map {
+    val servicePortIterator = templefile.services.zip(1024 until Int.MaxValue)
```
c-harding: This feels like something that should be done in the main code, and not in tests. See the suggested code snippet in the discussion tab (it was attached to outdated code)

Lewiky:
Boy did this get out of hand

c-harding (approved):

src/e2e/scala/temple/SimpleE2ETest.scala
```diff
@@ -43,5 +43,65 @@ class SimpleE2ETest extends FlatSpec with Matchers {
...
+    val templeKongKubeService = Files.readString(basePath.resolve("kube/kong").resolve("kong-service.yaml"))
+    templeKongKubeService shouldBe FileUtils.readResources("kube/kong/kong-service.yaml")
+
+    val templeKongKubeDbService = Files.readString(basePath.resolve("kube/kong").resolve("kong-db-service.yaml"))
+    templeKongKubeDbService shouldBe FileUtils.readResources("kube/kong/kong-db-service.yaml")
+
+    val templeKongKubeMigration = Files.readString(basePath.resolve("kube/kong").resolve("kong-migration-job.yaml"))
+    templeKongKubeMigration shouldBe FileUtils.readResources("kube/kong/kong-migration-job.yaml")
+
+    // Only these files should be present in the kube/TempleUser folder
```
jaylees14: We may want to lower case the `TempleUser`, to be consistent with the service folder name?

src/main/scala/temple/builder/OrchestrationBuilder.scala
```diff
@@ -0,0 +1,41 @@
...
+object OrchestrationBuilder {
+
+  def createServiceOrchestrationRoot(
+    projectName: String,
+    services: Seq[(String, ServiceBlock, Int)],
+  ): OrchestrationRoot =
+    OrchestrationRoot(
+      services map {
+        case (name, service, port) =>
+          val dockerImage = s"temple-$projectName-$name"
```
jaylees14: Not sure I like adding `temple` to the start of each Docker image - feels like it ties it too closely with generated code? @xsanda @smithwjv?

src/main/scala/temple/builder/project/ProjectBuilder.scala
```diff
@@ -28,13 +29,19 @@ object ProjectBuilder {
...
-      case ((name, service), port) =>
+    val dockerfiles = templefile.servicesWithPorts.map {
+      case (name, service, port) =>
         val dockerfileRoot     = DockerfileBuilder.createServiceDockerfile(name.toLowerCase, service, port)
         val dockerfileContents = DockerfileGenerator.generate(dockerfileRoot)
         (File(s"${name.toLowerCase}", "Dockerfile"), dockerfileContents)
     }
 
-    Project(databaseCreationScripts ++ dockerfiles)
+    val orchestrationRoot = OrchestrationBuilder.createServiceOrchestrationRoot(
```
jaylees14, line 40: Just a consistency side note - in the `database` and `dockerfile` cases we map over the services in this file, whereas with Kubernetes we map over in the builder - any particular reason why?

src/main/scala/temple/generate/kube/KubernetesGenerator.scala
jaylees14: Can we not also use `.map(EnvVar.tupled)`?

src/e2e/scala/temple/SimpleE2ETest.scala
jaylees14: I realise this is quite a few changes 😆, but we should probably be consistent

jaylees14 (commented):
Just a few more comments, but overall looks good. LMK if you don't agree

src/main/scala/temple/builder/OrchestrationBuilder.scala
c-harding: Is there a reason we didn’t just go for `s"$projectName-$name"`? @Lewiky @jaylees14

src/main/scala/temple/builder/project/ProjectBuilder.scala
Lewiky: The kube gen acts on all services at once (This way shared file i.e kong get generated properly)

jaylees14 (approved):
I'll just reiterate what we said in person: I don't really agree with Kubernetes being a single generator - I think it should be a per service thing, with the kong configuration added on top. This way it feels more modular, in the sense that if we wanted to switch API gateways in the future, it would be a single generator change, rather than having to dig through the entirety of the Kube generator.

PR #146, opened 2020-03-20
[grafana] Generate base grafana JSON

jaylees14:
Generate a base Grafana JSON, matching https://grafana.com/docs/grafana/latest/reference/dashboard/ and https://github.com/TempleEight/spec-golang/blob/develop/grafana/provisioning/dashboards/auth.json.

The only thing that changes for each config is the `panels`, which will come next :)


Lewiky (approved):
LGTM

c-harding (approved):

PR #148, opened 2020-03-20
[util] Move verb into the CRUD type

jaylees14:
It'll be useful to have a string representation of the `CRUD` value in the Grafana gen, so I've moved the function into an attribute on the type

### DIFF
--- a/src/main/scala/temple/generate/CRUD.scala
+++ b/src/main/scala/temple/generate/CRUD.scala
@@ -1,13 +1,13 @@
 package temple.generate
 
-sealed trait CRUD
+abstract class CRUD(val verb: String)
 
 object CRUD {
-  case object ReadAll extends CRUD
-  case object Create  extends CRUD
-  case object Read    extends CRUD
-  case object Update  extends CRUD
-  case object Delete  extends CRUD
+  case object ReadAll extends CRUD("List")
+  case object Create  extends CRUD("Create")
+  case object Read    extends CRUD("Read")
+  case object Update  extends CRUD("Update")
+  case object Delete  extends CRUD("Delete")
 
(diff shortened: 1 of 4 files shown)


src/main/scala/temple/generate/CRUD.scala
```diff
@@ -1,13 +1,13 @@
...
-  case object ReadAll extends CRUD
-  case object Create  extends CRUD
-  case object Read    extends CRUD
-  case object Update  extends CRUD
-  case object Delete  extends CRUD
+  case object ReadAll extends CRUD("List")
+  case object Create  extends CRUD("Create")
+  case object Read    extends CRUD("Read")
+  case object Update  extends CRUD("Update")
+  case object Delete  extends CRUD("Delete")
```
Lewiky, line 10: Pedantic thought - we've done this elsewhere by keeping the base type as a sealed trait and then having a toString defined on each of the implementations - which do we like better? @xsanda ?

Lewiky (approved):
LGTM

src/main/scala/temple/generate/CRUD.scala
jaylees14: Personally went for this method as `.verb`gives a little more semantic meaning than `.toString`, which could produce a debug description or something else unexpected

c-harding: This raises they question of why we have the difference between List and ReadAll. If we merge them then toString should be fair game

jaylees14: Think changing that is a little out of scope for this PR

c-harding (approved):

PR #156, opened 2020-03-23
[Semantics] Add Templefile validator

c-harding:
I’ll add a full description of side effects here after lunch, but main point is that there is now a validator, and it gets tested.

- Also includes side effects of updating the parser to support the new auth


src/main/scala/temple/DSL/semantics/Validator.scala
```diff
@@ -0,0 +1,96 @@
...
+      case _: Metadata.TargetLanguage    => // Nothing to validate yet
+      case Metadata.TargetAuth(_)        => context.fail(s"TODO: figure out how this auth works")
+      case _: Metadata.ServiceLanguage   => // Nothing to validate yet
+      case _: Metadata.Database          => // Nothing to validate yet
+      case _: Metadata.ServiceAuth       => // Nothing to validate yet
+      case _: Metadata.Readable          => // Nothing to validate yet
+      case _: Metadata.Writable          => // Nothing to validate yet
+      case Metadata.Omit(_)              => // Nothing to validate yet
+      case Metadata.ServiceEnumerable(_) => // Nothing to validate yet
+      case _: Metadata.Provider          => // Nothing to validate yet
```
Lewiky: When would there be validation done here?

Lewiky (approved):
LGTM my dude

src/main/scala/temple/DSL/semantics/Validator.scala
c-harding: Good question, do any of these need validation?

src/e2e/scala/temple/DSL/ParserE2ETest.scala
```diff
@@ -41,20 +41,22 @@ class ParserE2ETest extends FlatSpec with Matchers with DSLParserMatchers {
...
-                "friend" -> Attribute(ForeignKey("User")),
+                "friend" -> Attribute(ForeignKey("TempleUser")),
                 "image"  -> Attribute(BlobType(size = Some(10_000_000))),
               ),
               Seq(ServiceEnumerable(byThis = true), Readable.This),
             ),
           ),
         ),
+        "Booking" -> ServiceBlock(Map()),
+        "Event"   -> ServiceBlock(Map()),
```
jaylees14, line 57: Not that this is in scope for this PR, but I'm curious what would happen in all our generation if no fields were provided by the user

src/main/scala/temple/Application.scala
```diff
@@ -4,29 +4,29 @@ import temple.DSL.DSLProcessor
...
 
   def generate(config: TempleConfig): Unit = {
     val outputDirectory = config.Generate.outputDirectory.getOrElse(System.getProperty("user.dir"))
     val fileContents    = FileUtils.readFile(config.Generate.filename())
-    DSLProcessor.parse(fileContents) match {
-      case Left(error) => throw new RuntimeException(error)
-      case Right(data) =>
-        val analyzedTemplefile = Analyzer.parseSemantics(data)
-        val project            = ProjectBuilder.build(analyzedTemplefile)
+    val data = DSLProcessor.parse(fileContents) fromEither { error =>
```
jaylees14, line 14: Oooh nice

src/main/scala/temple/DSL/semantics/Validator.scala
jaylees14: Can't think of any - except for the one you've written below. I'm happy to leave this like this though, means when we think of things to add it's easy to do so

src/main/scala/temple/DSL/semantics/Validator.scala
```diff
@@ -0,0 +1,96 @@
...
+      case Metadata.Uses(services) =>
+        for (service <- services if !allServices.contains(service)) context.fail(s"No such service $service")
+    }
+
+  private def validateAttribute(attribute: Attribute)(context: SemanticContext): Unit = {
+    validateAttributeType(attribute.attributeType)(context)
+    attribute.accessAnnotation match {
+      case Some(Annotation.Client)    => // nothing to validate
+      case Some(Annotation.Server)    => // nothing to validate
+      case Some(Annotation.ServerSet) => // nothing to validate
```
jaylees14, line 71: I'm wondering, do we validate that these are exclusive - i.e would it ever make sense to be both `@client` and `@server`

src/main/scala/temple/DSL/semantics/Validator.scala
```diff
@@ -0,0 +1,96 @@
...
+      case None                       => // nothing to validate
+    }
+  }
+
+  private def validateAttributeType(attributeType: AttributeType)(context: SemanticContext): Unit =
+    attributeType match {
+      case ForeignKey(references) if !allStructs.contains(references) =>
```
jaylees14: If I read this right, this allows you to have:

```
Foo: service {...}

Bar: service {
    ...
    Baz: struct {
      foo: Foo
    }
}
```

I haven't put much though into exactly what that means, but are we ok to support that? Since at that stage they're not technically foreign keys because we can't implement that constraint over 2 databases. @smithwjv thoughts?

src/main/scala/temple/DSL/syntax/Arg.scala
```diff
@@ -1,12 +1,22 @@
 package temple.DSL.syntax
 
 /** A value that can be passed as an argument to a type or metadata */
-sealed trait Arg
+sealed trait Arg {
+  def toTempleString: String
```
jaylees14, line 5: <3

jaylees14 (approved):
Nice stuff! Just a couple of comments that popped into my head as I was reading, but code looks good :)

src/main/scala/temple/DSL/semantics/Validator.scala
c-harding: They are mutually exclusive, in that there’s only one option variable in which it can be stored. Adding two would be an error: I can add a test for that if you’d like?

src/main/scala/temple/DSL/semantics/Validator.scala
c-harding: And likewise `Foo.field` can reference `Baz`. Is this desirable?

src/main/scala/temple/DSL/semantics/Validator.scala
jaylees14: Ah I see makes sense, all good :)

src/main/scala/temple/DSL/semantics/Validator.scala
jaylees14: Let's leave it like that for now, and restrict it later if we need to. I don't think we've actually got an example of that in Go, so let's try and work one out in our examples and circle back on this

PR #157, opened 2020-03-23
[go-gen] Refactor dao generation

willjvsmith:
* Removes exceptions through CreatedBy enumeration enum
  * Forces desired state
* Updates tests

I've learnt more Scala than I wanted to


src/main/scala/temple/generate/server/CreatedByAttribute.scala
```diff
@@ -2,8 +2,38 @@ package temple.generate.server
...
+sealed trait CreatedByAttribute
+
+object CreatedByAttribute {
+
+  /**
+    * None encapsulates the case where no Temple defined createdBy field exists, e.g. in the case of an Auth block
+    */
+  case object None extends CreatedByAttribute
+
+  sealed trait Enumerating extends CreatedByAttribute{
```
c-harding: Please run the autoformatter over everything
```suggestion
  sealed trait Enumerating extends CreatedByAttribute {
```

jaylees14:
```suggestion
  sealed trait Enumerating extends CreatedByAttribute {
```

src/main/scala/temple/generate/server/go/service/GoServiceDAOGenerator.scala
```diff
@@ -51,10 +51,14 @@ object GoServiceDAOGenerator {
   private def generateDatastoreInterfaceFunction(
     serviceName: String,
     operation: CRUD,
-    enumByCreatedBy: Boolean,
+    createdByAttribute: CreatedByAttribute,
   ): String = {
     val functionName = generateDAOFunctionName(operation, serviceName)
-    val functionArgs = if (enumByCreatedBy || operation != CRUD.ReadAll) s"input ${functionName}Input" else ""
+    val enumeratingByCreator = createdByAttribute match {
```
c-harding, line 57: Is there a reason why this isn’t `createdByAttribute.isInstanceOf[CreatedByAttribute.EnumerateByCreator]`?

jaylees14 (approved):
LGTM

src/main/scala/temple/generate/server/CreatedByAttribute.scala
```diff
@@ -2,8 +2,38 @@ package temple.generate.server
 
 import temple.ast.AttributeType
 
-case class CreatedByAttribute(
-  inputName: String,
-  name: String,
-  attributeType: AttributeType,
-)
+sealed trait CreatedByAttribute
```
c-harding, line 5: I know this has taken quite a while, but it feels like `Option[CreatedByAttribute]` (where `case class CreatedByAttribute(inputName: String, name: String, attributeType: AttributeType, filterEnumeration: Boolean)`, would be better, given that

1) There’s no difference between the arguments of `EnumerateByCreator` and `EnumerateByAll`, either in storage or semantics.
2) The switching between them is only ever done in a boolean manner.
3) CreatedByAttribute.None is serving the exact purpose of Option.None.

Would be interested to hear @jaylees14’s thoughts too

jaylees14: Yeah I'm okay with that 👍

src/main/scala/temple/generate/server/go/service/GoServiceDAOGenerator.scala
willjvsmith: Discussed offline, essentially isInstanceOf uses reflection under the hood, and not pattern matching for some reason

src/main/scala/temple/generate/server/CreatedByAttribute.scala
willjvsmith: The reason we went with this is because we could feasibly add different enumerations in the future

willjvsmith: Discussed offline, it wasn't worth the time to switch

c-harding (approved):

Lewiky (approved):
LGTM

PR #159, opened 2020-03-23
[go-gen] Add dao init and query function generation

willjvsmith:
* Adds generation of DAO init function
* Adds generation of DAO query functions
* Updates tests
  * Removes comments from imports to be generated

Doesn't call to read from file, just for you @jaylees14!
Next PR will generate the CRUD DAO functions, thought I'd keep the PRs separate.


jaylees14 (approved):
LGTM

c-harding (approved):
nice

PR #160, opened 2020-03-23
[grafana] Generate yml config for grafana

jaylees14:
On top of the beautiful JSON, Grafana also needs a couple of `.yml` config files.

This is the first (from https://github.com/TempleEight/spec-golang/blob/develop/grafana/provisioning/dashboards/dashboards.yml), which is always the same for our datasource, but I've taken the name as an argument so we can be consistent everywhere

### DIFF
--- a/src/test/scala/temple/generate/metrics/grafana/GrafanaDashboardConfigGeneratorTestData.scala
+++ b/src/test/scala/temple/generate/metrics/grafana/GrafanaDashboardConfigGeneratorTestData.scala
@@ -3,7 +3,7 @@ package temple.generate.metrics.grafana
 object GrafanaDashboardConfigGeneratorTestData {
 
   val prometheusConfig: String =
-    """|apiVersion: 1
+    """apiVersion: 1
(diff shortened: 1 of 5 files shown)


src/test/scala/temple/generate/metrics/grafana/GrafanaDashboardConfigGeneratorTestData.scala
```diff
@@ -0,0 +1,18 @@
+package temple.generate.metrics.grafana
+
+object GrafanaDashboardConfigGeneratorTestData {
+
+  val prometheusConfig: String =
+    """|apiVersion: 1
```
c-harding:
```suggestion
    """apiVersion: 1
```

c-harding (approved):
Not sure how your multi-line comment system is set up, but it looks wrong

willjvsmith (approved):
LGTM 🚀

Lewiky (approved):
Easy easy

PR #161, opened 2020-03-23
[gen] Rename ReadAll to List for consistency

c-harding:
Could add the following, to allow for `.verb` instead of `.toString`, if Jay’s Swifty ways would prefer.

```scala
sealed trait CRUD {
  def verb: String = toString
}
```


jaylees14 (approved):
LGTM

willjvsmith (commented):
Whoops reviewed #162 first

Nonetheless, some function names in the generator refer to `ReadAll`, probs want changing too

willjvsmith (approved):
Is going to cause conflicts with my current un-PRed work, but I'll fix them when I PR

PR #162, opened 2020-03-23
[gen] Make CRUD a simple enum

c-harding:
Stacks on #161, using native enums so we can use sorting on collections of endpoints


src/main/scala/temple/generate/CRUD.scala
```diff
@@ -1,13 +1,6 @@
...
-object CRUD {
-  case object ReadAll extends CRUD("List")
-  case object Create  extends CRUD("Create")
-  case object Read    extends CRUD("Read")
-  case object Update  extends CRUD("Update")
-  case object Delete  extends CRUD("Delete")
-
-  val values = Seq(ReadAll, Create, Read, Update, Delete)
+object CRUD extends Enumeration {
+  type CRUD = Value
```
jaylees14, line 4: Where for art thou value defined?

jaylees14 (approved):
LGTM, indifferent if we change or not, agree does read a little nicer

src/main/scala/temple/builder/MetricsBuilder.scala
```diff
@@ -1,6 +1,6 @@
 package temple.builder
 
-import temple.generate.CRUD
+import temple.generate.CRUD.CRUD
```
willjvsmith, line 3: nice

willjvsmith (commented):
Some of the functions in the Go generators refer to `ReadAll` in their names, probs want changing too

src/main/scala/temple/generate/CRUD.scala
c-harding: @jaylees14 it’s a field of the Enumeration class: https://www.scala-lang.org/api/current/scala/Enumeration.html

willjvsmith:
Will approve once updated and conflicts resolved

willjvsmith (approved):

PR #163, opened 2020-03-25
[grafana] Output dashboard config

jaylees14:
Additionally output the configuration file for the `grafana` dashboards, which sits at the same level as the `.json` files 

(https://github.com/TempleEight/spec-golang/blob/develop/grafana/provisioning/dashboards)


Lewiky (approved):
LGTM bud

willjvsmith (approved):
LGTM 🚀

PR #164, opened 2020-03-25
[grafana] Use custom type for datasource

jaylees14:
When creating a datasource in Grafana, the user can assign it a name. This is then what's used to reference it from panels etc.

At the moment I was just using strings for this, but it feels nicer to wrap it in it's own type and then use this everywhere. 

The fact we're using "Prometheus" for the name is somewhat coincidental, we can use any unique identifier here.


willjvsmith (approved):
LGTM 🚀

Lewiky (approved):
LGTM

PR #165, opened 2020-03-25
[Application] Implement question asking functionality and details

Lewiky:
Adds the logic for getting extra language-specific generation details from the user at runtime, things that aren't mentioned in the Templefile like the go module path.


src/main/scala/temple/detail/LanguageSpecificDetailBuilder.scala
```diff
@@ -0,0 +1,26 @@
...
+import temple.ast.{Metadata, Templefile}
+import temple.builder.project.ProjectConfig
+import temple.detail.Detail.GoDetail
+
+/** Call out to the user and ask for any required extra input based on their templefile options */
+object LanguageSpecificDetailBuilder {
+
+  private def buildGoDetail(templefile: Templefile, questionAsker: QuestionAsker): Detail = {
+    val packageName = questionAsker.askQuestion(
+      "Yo pal, what is the Go module name you'd like (expected format \"github.com/yeeter69/lockdown\")?",
```
jaylees14: Probably should update this, since it's user facing 😆

src/main/scala/temple/detail/LanguageSpecificDetailBuilder.scala
```diff
@@ -0,0 +1,26 @@
+package temple.detail
+
+import temple.ast.Metadata.ServiceLanguage
+import temple.ast.{Metadata, Templefile}
+import temple.builder.project.ProjectConfig
+import temple.detail.Detail.GoDetail
+
+/** Call out to the user and ask for any required extra input based on their templefile options */
```
jaylees14:
```suggestion
/** Call out to the user and ask for any required extra input based on their Templefile options */
```

src/main/scala/temple/detail/LanguageSpecificDetailBuilder.scala
```diff
@@ -0,0 +1,26 @@
...
+object LanguageSpecificDetailBuilder {
+
+  private def buildGoDetail(templefile: Templefile, questionAsker: QuestionAsker): Detail = {
+    val packageName = questionAsker.askQuestion(
+      "Yo pal, what is the Go module name you'd like (expected format \"github.com/yeeter69/lockdown\")?",
+    )
+    GoDetail(packageName)
+  }
+
+  /** Given a templefile and a question asker, build all the required details */
```
jaylees14:
```suggestion
  /** Given a Templefile and a QuestionAsker, build the additional details required to generate */
```

src/main/scala/temple/detail/Detail.scala
```diff
@@ -0,0 +1,15 @@
+package temple.detail
+
+/** Maintains any extra required information the user did not provide in the Templefile */
+sealed trait Detail
```
jaylees14: I'd be tempted to call this `LanguageDetail` just to be more specific

src/main/scala/temple/detail/LanguageSpecificDetailBuilder.scala
```diff
@@ -0,0 +1,26 @@
...
+import temple.builder.project.ProjectConfig
+import temple.detail.Detail.GoDetail
+
+/** Call out to the user and ask for any required extra input based on their templefile options */
+object LanguageSpecificDetailBuilder {
+
+  private def buildGoDetail(templefile: Templefile, questionAsker: QuestionAsker): Detail = {
+    val packageName = questionAsker.askQuestion(
+      "Yo pal, what is the Go module name you'd like (expected format \"github.com/yeeter69/lockdown\")?",
+    )
```
jaylees14, line 14: What happens if the user returns the empty string? Surely we want some validation here?

jaylees14 (commented):
LGTM, just a few comment to tidy it up

src/main/scala/temple/detail/LanguageSpecificDetailBuilder.scala
```diff
@@ -0,0 +1,32 @@
...
+
+import temple.ast.Metadata.ServiceLanguage
+import temple.ast.{Metadata, Templefile}
+import temple.builder.project.ProjectConfig
+import temple.detail.LanguageDetail.GoLanguageDetail
+
+/** Call out to the user and ask for any required extra input based on their Templefile options */
+object LanguageSpecificDetailBuilder {
+
+  private def buildGoDetail(templefile: Templefile, questionAsker: QuestionAsker): LanguageDetail = {
```
c-harding: This is specific to Go, is there a reason it doesn’t return type `GoLanguageDetail`?

src/main/scala/temple/detail/LanguageDetail.scala
```diff
@@ -0,0 +1,15 @@
+package temple.detail
+
+/** Maintains any extra required information the user did not provide in the Templefile */
+sealed trait LanguageDetail
```
c-harding, line 4: Must say I’m a little confused as to how this trait is going to be used

Lewiky: The idea is that it gets passed with the Templefile into the ProjectBuilder, and can then get used to gain the extra detail needed

c-harding: I meant more that it seems orthogonal to the directions in which classes should go: if you call build, you get a LanguageDetail, but there’s no way to do anything with a LanguageDetail by itself other than pattern matching, when that pattern matching is parallel to the language itself

c-harding: I’m too tired to think of a solution now, but it seems like a bit of a strange use of the type system

c-harding: I’ll accept this for now, but may suggest something more when the PR using it comes in

c-harding (approved):

jaylees14 (approved):
LGTM 👍

PR #167, opened 2020-03-25
[go-gen] Add CodeTermNone

c-harding:
Not tested (may be a solution for your problem of expecting codeterm, @smithwjv)

### DIFF
--- a/src/main/scala/temple/generate/utils/CodeTerm.scala
+++ b/src/main/scala/temple/generate/utils/CodeTerm.scala
@@ -20,10 +20,14 @@ trait CodeTerm {
 
 object CodeTerm {
 
-  /** Turns a [[String]] into a [[CodeTerm]], with a trivial iterator of just the single string
-    */
+  /** Turns a [[String]] into a [[CodeTerm]], with a trivial iterator of just the single string */
   implicit class CodeTermString(string: String) extends CodeTerm {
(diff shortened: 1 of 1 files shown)


willjvsmith:
Ok nice, I'll give it a go and see if it works before approving

willjvsmith (approved):
Yep does the trick, I was casting to Option[String], but if we get this in I can merge into my branch

src/main/scala/temple/generate/utils/CodeTerm.scala
```diff
@@ -20,12 +20,16 @@ trait CodeTerm {
...
 
-  /** Turns a [[String]] into a [[CodeTerm]], with a trivial iterator of just the single string
-    */
+  /** Turns a [[String]] into a [[CodeTerm]], with a trivial iterator of just the single string */
   implicit class CodeTermString(string: String) extends CodeTerm {
     override def flatIterator: Iterator[String] = Iterable.single(string).iterator
   }
 
+  /** Turns a [[None]] into a [[CodeTerm]], with a trivial iterator of nothing */
+  implicit class CodeTermNone(none: None.type) extends CodeTerm {
```
jaylees14, line 29: Can we not use `Optional` here, flatten out the `None`? Feels like it could be useful - similar to the JSON stuff

jaylees14 (approved):
LGTM

PR #169, opened 2020-03-25
[go-gen] Add dao list interface function generation

willjvsmith:
* Adds generation of the `List` DAO interface function
* Adds generation of comments and declarations for all DAO interface functions
* Combines CRUD operations and corresponding queries in `ServiceRoot` field `opQueries`
* Adds datastore parameter to `ServiceRoot`
  * Currently unused, but will be used to switch on datastore imports/functions
* Updates tests
  * Returns nil in unfinished functions to make integration tests pass

Ignore .snippets, meat is in `GoServiceDAOGenerator`


src/it/scala/temple/generate/server/GoGeneratorIntegrationTestData.scala
```diff
@@ -11,26 +12,38 @@ object GoGeneratorIntegrationTestData {
     "user",
     "github.com/TempleEight/spec-golang/user",
     Seq.empty,
-    Set(CRUD.Create, CRUD.Read, CRUD.Update, CRUD.Delete),
+    ListMap(
+      CRUD.Create -> "INSERT INTO user_temple (id, name) VALUES ($1, $2) RETURNING id, name",
+      CRUD.Read   -> "SELECT id, name FROM user_temple WHERE id = $1",
+      CRUD.Update -> "UPDATE user_temple SET name = $1 WHERE id = $2 RETURNING id, name",
+      CRUD.Delete -> "DELETE FROM user_temple WHERE id = $1",
```
jaylees14, line 19: I'd be tempted to make these strings come from the Postgres generator - especially given this is an integration test. Thoughts?

src/main/scala/temple/generate/database/DatabaseGenerator.scala
```diff
@@ -3,7 +3,8 @@ package temple.generate.database
 import temple.generate.database.ast.Statement
 
 /** DatabaseGenerator provides an interface for generating DB specific query languages from an AST */
-trait DatabaseGenerator[Context <: DatabaseContext] {
+trait DatabaseGenerator {
+  type Context <: DatabaseContext
```
jaylees14, line 7: Why?

src/main/scala/temple/generate/server/go/service/GoServiceDAOGenerator.scala
```diff
@@ -191,6 +192,148 @@ object GoServiceDAOGenerator {
...
+      },
+    )
+
+  private def generateListInterfaceFunctionBodyQueryBlock(
+    createdByAttribute: CreatedByAttribute,
+    query: String,
+  ): String =
+    mkCode.lines(
+      CodeWrap.parens.prefix("rows, err := executeQueryWithRowResponses")(
+        mkCode.list(
```
jaylees14: I'm sure we've spoken about this before, but it really feels like we're hitting the point of making some Go utility functions that abstract away the mkCode logic a bit

```scala
callGoFunction(returnVars = "rows, err", function = "executeQueryWithRowResponses", args)
```

src/main/scala/temple/generate/server/go/service/GoServiceDAOGenerator.scala
```diff
@@ -191,6 +192,148 @@ object GoServiceDAOGenerator {
...
+            case _                                   => ""
+          },
+        ),
+      ),
+      mkCode(
+        "if err != nil",
+        CodeWrap.curly.tabbed(
+          "return nil, err",
+        ),
+      ),
```
jaylees14: This should definitely be a util, it's used everywhere - `checkAndReturnError`

src/main/scala/temple/generate/server/go/service/GoServiceDAOGenerator.scala
```diff
@@ -191,6 +192,148 @@ object GoServiceDAOGenerator {
...
+    attributes: ListMap[String, Attribute],
+  ): String =
+    mkCode.lines(
+      s"${serviceName}List := make([]${serviceName.capitalize}, 0)",
+      mkCode(
+        "for rows.Next()",
+        CodeWrap.curly.tabbed(
+          s"var $serviceName ${serviceName.capitalize}",
+          CodeWrap.parens.prefix("err = rows.Scan")(
+            mkCode.list(
```
jaylees14: Some variables here might be nice?

jaylees14 (commented):
Changes look good. I'm really thinking it's time to introduce some helper functions, it's kinda exploding. Also wondering if there's a way to split up the functions into separate files - I understand they are all for the DAO, but it's getting kinda big and you've only done `list`

src/main/scala/temple/generate/database/DatabaseGenerator.scala
willjvsmith: @xsanda 

This comes from previously wanting to pass a `PostgresGenerator` around to be used inside the Go generation to generate Postgres. Decided against that approach, but Charlie thought this should still be like this.

src/it/scala/temple/generate/server/GoGeneratorIntegrationTestData.scala
willjvsmith: Yeah fair good shout

willjvsmith: Unfortunately our Postgres gen doesn't support a couple things needed for this, so I have left them as strings ready to be updated when the necessary functionality is added to PostgresGenerator.

src/main/scala/temple/generate/server/go/service/GoServiceDAOGenerator.scala
willjvsmith: Think I'll look into this in a new PR, where I refactor GoServiceDAOGenerator into separate files too.

Lewiky (approved):
LGTM - Agree with Jay's comments

src/main/scala/temple/generate/database/DatabaseGenerator.scala
c-harding: The correct context needs to be passed around for the correct generator, and having this as a parameter wasn’t as flexible as having it as a type argument (there was some weird contravariant type error)

c-harding: It means there is a type called `Context`, defined in the concrete implementations, which is a subtype of `DatabaseContext`.

willjvsmith:
I will refactor DAOGenerator into separate files in a separate PR

jaylees14 (approved):
LGTM, think you're missing some cases in the unit tests (where `createdBy` is `None`), but happy to go as is

src/main/scala/temple/generate/database/DatabaseGenerator.scala
jaylees14: Seems familiar https://www.hackingwithswift.com/articles/74/understanding-protocol-associated-types-and-their-constraints :P

PR #172, opened 2020-03-25
[go-gen] Refactor dao generation

willjvsmith:
* Splits up what was a ~350 line file into smaller ones
  * Breaks the pattern, but makes it more manageable to see what's going on to an unfamiliar

Recommend pulling and taking a look at the structure, also I'm not so happy with the naming of functions/files, so if u think of better ones feel free to comment


c-harding (approved):
looks sensible

Lewiky (approved):
Like it

jaylees14 (approved):
Yeah as we spoke about offline, some of the names are getting quite long, but I think it's worth the compromise

PR #174, opened 2020-03-26
[sql-gen] Support prepared statements in update and where

jaylees14:
Sorry for the PR size, quite a lot of changes were needed.

We now additionally support queries of the following structure:
- `UPDATE table SET value1 = $1, value2 = $2` - Prepared updates 
- `UPDATE table SET value1 = 5, value2 = $1` - Mixed prepared and provided updates
- `SELECT * FROM table WHERE value1 = $1` - Prepared WHERE clauses 

Also had to make sure that when we're doing `Disjunction` and `Conjunction` that the dollars are provided in the correct order, so have added some tests for that.

The final thing to support is mixed prepared and provided values in `INSERT` statements, but I'll do that in a follow up PR since this one was getting big. I'll also add all of the queries from `spec-golang` as unit tests in that PR to make sure we hit everything we need

### DIFF
--- a/src/main/scala/temple/generate/database/PostgresGenerator.scala
+++ b/src/main/scala/temple/generate/database/PostgresGenerator.scala
@@ -12,35 +12,32 @@ import temple.generate.utils.CodeTerm._
 import scala.Option.when
 
 /** Implementation of [[DatabaseGenerator]] for generating PostgreSQL */
-object PostgresGenerator extends DatabaseGenerator {
-  type Context = PostgresContext
+private class PostgresGenerator private (context: PostgresContext) {
+
+  private var currentIndex = 1
+
+  private def nextPreparedSymbol(): String = {
+    val symbol = context.preparedType match {
+      case PreparedType.QuestionMarks => "?"
+      case PreparedType.DollarNumbers => "$" + currentIndex
+    }
+    currentIndex += 1
+    symbol
+  }
 
   /** Given an expression, parse it into the Postgres format */
-  private def generateExpression(expression: Expression, index: Int)(implicit context: PostgresContext): (String, Int) =
+  private def generateExpression(expression: Expression): String =
     expression match {
-      case Value(value) =>
-        (value, index)
-      case PreparedValue =>
-        val value = context.preparedType match {
-          case PreparedType.QuestionMarks => "?"
-          case PreparedType.DollarNumbers => "$" + index
-        }
-        (value, index + 1)
+      case Value(value)  => value
+      case PreparedValue => nextPreparedSymbol()
     }
 
   /** Given a list of assignments, parse it into the Postgres format */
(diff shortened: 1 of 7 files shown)


src/main/scala/temple/generate/database/PostgresGenerator.scala
```diff
@@ -16,14 +16,31 @@ object PostgresGenerator extends DatabaseGenerator {
   type Context = PostgresContext
 
   /** Given an expression, parse it into the Postgres format */
-  private def generateExpression(expression: Expression): String =
+  private def generateExpression(expression: Expression, index: Int)(implicit context: PostgresContext): (String, Int) =
```
xsanda: This feels like rather a lot of awkward threading of the position int through here, especially when the context is passed around too. Could the context instead expose a `nextPreparedSymbol` method? This would mean that a fresh PostgresContext would be needed for each query to be generated. This could be achieved by

1. Replacing `implicit val context` with `implicit def context` everywhere outside this class
1. ContextFactory, though for stateless database contexts this would be a trivial implementation of `class MyDBContext extends DBContextFactory with DBContext {def make = this}`

Another way to do it could be to make a `PostgresGenerator` class, with this index as a variable

c-harding: This feels like rather a lot of awkward threading of the position int through here, especially when the context is passed around too. Could the context instead expose a `nextPreparedSymbol` method? This would mean that a fresh PostgresContext would be needed for each query to be generated. This could be achieved by one of

1. Replacing `implicit val context` with `implicit def context` everywhere outside this class
1. ContextFactory, though for stateless database contexts this would be a trivial implementation of `class MyDBContext extends DBContextFactory with DBContext {def make = this}`

Another way to do it could be to make a `PostgresGenerator` class, with this index as a property

Lewiky (commented):
I agree with Charlie here, it's kinda hard to follow what's going on with the index and know what it's value is at different points and where it's reinitialized and where it's incremented, hopefully putting it into the context makes it a little clearer

src/main/scala/temple/generate/database/PostgresGenerator.scala
jaylees14: We chatted about this offline - going to go for the `class` method, which can be abstracted from the caller, rather than expecting them to update

Lewiky (approved):
Yeah this is easier to follow 👍

c-harding (approved):
Looks very nice

src/main/scala/temple/generate/database/PostgresGenerator.scala
c-harding: "offline"
Our whole world is online, we could literally be on our own planets right now with just the internet connecting us

PR #175, opened 2020-03-26
[sql-gen] Allow arguments to be preprovided to the insert query

jaylees14:
Support INSERT statements where values are pre-provided

Also add unit tests for each query we make in `spec-golang`, so we know it works :P


c-harding (approved):
lgtm

willjvsmith (approved):
Good stuff

PR #170, opened 2020-03-25
[grafana] Generate datasource YML for Grafana

jaylees14:
Generate the next `yml` file for Grafana: `datasources.yml`

See
https://github.com/TempleEight/spec-golang/blob/develop/grafana/provisioning/datasources/datasource.yml

PR #176, opened 2020-03-26
[grafana] Add datasource config to mapping

jaylees14:
Add yml file generated in #170 to the mapping


src/main/scala/temple/builder/project/ProjectBuilder.scala
```diff
@@ -72,9 +72,11 @@ object ProjectBuilder {
           val rows             = MetricsBuilder.createDashboardRows(name, datasource, endpoints(service))
           val grafanaDashboard = GrafanaDashboardGenerator.generate(name.toLowerCase, name, rows)
           (File(s"grafana/provisioning/dashboards", s"${name.toLowerCase}.json"), grafanaDashboard)
-      } + (
+      } ++ Map(
```
jaylees14, line 75: I think I might abstract this into the metrics builder so it's just one call that returns all the files. Will do in a follow up tho

willjvsmith (approved):
Not familiar with the semantics of what you're generating, but generation LGTM 🚀

c-harding (approved):
Ditto Will, looks sensible

PR #177, opened 2020-03-26
[go-gen] Generate dao function bodies

willjvsmith:
* Adds generation of dao function bodies
* Adds helper functions to make generation code more readable
* Adds `noIndent` and `noIndentList` functions to CodeTerm
  * Same behaviour as spaced and tabbed versions, but without the spaces or tabs 💯 

This completes `dao.go` for non-auth services, yay! I will however do a slight refactor in the next PR to pass around the serviceRoot rather than passing around it's constituent parts. This will make the file smaller. As usual ignore `.snippet`s.

Note I will add a helper function for 'if error switch ' blocks when I come to the handlers where they're also used.


src/main/scala/temple/generate/server/go/common/GoCommonGenerator.scala
```diff
@@ -42,4 +42,43 @@ object GoCommonGenerator {
         ),
       ),
     )
+
+  /** Generate declaration and assignment statement */
+  private[go] def genDeclareAndAssign(expr: String, identifiers: String*): String =
+    mkCode.list(identifiers) + " := " + expr // mkCode doesn't put a space before the colon
```
c-harding, line 48: Could do `mkCode(mkCode.list(identifiers), " :=", expr)`, but you’re right, mkCode doesn’t add a space because it’s rare for colons to be preceded by a space. Not sure how to make it nicer though

src/main/scala/temple/generate/server/go/common/GoCommonGenerator.scala
```diff
@@ -42,4 +42,43 @@ object GoCommonGenerator {
...
+
+  /** Generate variable declaration */
+  private[go] def genVar(identifier: String, typ: String): String =
+    mkCode(
+      "var",
+      identifier,
+      typ,
+    )
+
+  //** Generate function call */
```
jaylees14:
```suggestion
  /** Generate function call */
```

src/main/scala/temple/generate/server/go/common/GoCommonGenerator.scala
```diff
@@ -42,4 +42,43 @@ object GoCommonGenerator {
         ),
       ),
     )
+
+  /** Generate declaration and assignment statement */
+  private[go] def genDeclareAndAssign(expr: String, identifiers: String*): String =
```
jaylees14, line 47: I see why you've had to do this, but I wish you could switch the order of the args 😢

src/main/scala/temple/generate/server/go/service/dao/GoServiceDAOFunctionsGenerator.scala
```diff
@@ -42,6 +43,185 @@ object GoServiceDAOFunctionsGenerator {
...
+    }
+  }
+
+  private def generateQueryBlockErrorHandling(
+    serviceName: String,
+    operation: CRUD,
+    idAttribute: IDAttribute,
+  ): Option[String] =
+    operation match {
+      case List => Some(generateCheckAndReturnError("nil"))
```
jaylees14: Personal preference is that if one arrow goes onto many lines, they all should - up to you

src/main/scala/temple/generate/server/go/service/dao/GoServiceDAOFunctionsGenerator.scala
```diff
@@ -42,6 +43,185 @@ object GoServiceDAOFunctionsGenerator {
...
+      case Delete =>
+        Some(
+          mkCode(
+            generateCheckAndReturnError(),
+            mkCode(
```
c-harding: Is there a reason for a mkCode inside a mkCode?

jaylees14 (approved):
LGTM, much easier to read with those utility functions :)

src/main/scala/temple/generate/server/go/common/GoCommonGenerator.scala
willjvsmith: Yeah it makes me sad 😞

src/main/scala/temple/generate/server/go/service/dao/GoServiceDAOFunctionsGenerator.scala
willjvsmith: Wish the formatter was a bit more opinionated

jaylees14: Or I could stop being lazy and make the formatter do this for us.. I'll do that 😄

src/main/scala/temple/generate/server/go/common/GoCommonGenerator.scala
willjvsmith: I had that, but figured then the outer mkCode wasn't really doing it's job

willjvsmith: Don't think it's particularly necessary to fix

src/main/scala/temple/generate/server/go/common/GoCommonGenerator.scala
c-harding: Could curry it and reverse them, but up to you if you prefer that

src/main/scala/temple/generate/server/go/service/dao/GoServiceDAOFunctionsGenerator.scala
```diff
@@ -42,6 +43,185 @@ object GoServiceDAOFunctionsGenerator {
...
+                ),
+              ),
+              "default:",
+              tabIndent(
+                genReturn("nil", "err"),
+              ),
+            ),
+          ),
+        ),
+      ),
```
c-harding, line 198: This will definitely benefit from being factored out, but I guess exactly how will depend on how much variation there is. It looks to me like you’ve got something along the lines of `genSwitch(Some("sql.ErrorNoRows") -> genReturn(...), Nothing -> genReturn("nil", "err"))`

c-harding (approved):
Looks very good. How’s noIndentList going to be used?

src/main/scala/temple/generate/server/go/service/dao/GoServiceDAOFunctionsGenerator.scala
willjvsmith: Yeah there's a bit of variation on the switch expression itself too, e.g. switching on the type of the error as seen in the handlers of match/user.go

PR #178, opened 2020-03-26
[prom] Add prometheus config generator

jaylees14:
Add the final `.yml` file that needs to be generated for metrics: `prometheus.yml`

This PR generates it (essentially following https://github.com/TempleEight/spec-golang/blob/develop/prometheus/prometheus.yml), and the next will add the mapping.

I did have to change the generated file slightly, since the one prometheus gives as a template mixes styles of sequences, so I've stuck to one here.

### DIFF
--- a/src/main/scala/temple/generate/metrics/prometheus/PrometheusConfigGenerator.scala
+++ b/src/main/scala/temple/generate/metrics/prometheus/PrometheusConfigGenerator.scala
@@ -1,0 +1,12 @@
+package temple.generate.metrics.prometheus
+
+import io.circe.yaml.Printer
+import io.circe.syntax._
+import io.circe.yaml.Printer.FlowStyle
+import temple.generate.metrics.prometheus.ast.{PrometheusConfig, PrometheusJob}
+
+object PrometheusConfigGenerator {
+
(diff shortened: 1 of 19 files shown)


willjvsmith (approved):

src/main/scala/temple/generate/metrics/prometheus/PrometheusConfigGenerator.scala
```diff
@@ -0,0 +1,12 @@
+package temple.generate.metrics.prometheus
+
+import io.circe.yaml.Printer
+import io.circe.syntax._
+import io.circe.yaml.Printer.FlowStyle
```
c-harding: Ooh, what’s this?

c-harding (approved):
lgtm

src/main/scala/temple/generate/metrics/prometheus/PrometheusConfigGenerator.scala
jaylees14: Oops, this was me playing around with the parameters to the `Printer`. I'll run optimise imports 🙋‍♂

PR #179, opened 2020-03-26
[go-gen] Refactor to pass around service root rather than deconstructing

willjvsmith:
* Refactors to passing around the service root, rather than parts of it
  * Makes the code less verbose

Shout out to the Holy Church of the Squat and Dab who were relocated from their homeland this week


c-harding (approved):
Very nice

jaylees14 (approved):
Much nicer 👍

PR #180, opened 2020-03-26
[prom] Map into prometheus configuration generator

jaylees14:
Final one for the prometheus/grafana config - we now generate the `.yml` file from the types in #178, and pop it in the right place.

Nothing too crazy here, although thoughts on the `Ports` type?

### DIFF
--- a/src/e2e/scala/temple/SimpleE2ETestData.scala
+++ b/src/e2e/scala/temple/SimpleE2ETestData.scala
@@ -245,4 +245,15 @@ object SimpleE2ETestData {
       |  isDefault: true
       |  editable: true
       |""".stripMargin
+
+  val prometheusConfig: String =
+    """global:
+    |  scrape_interval: 15s
+    |  evaluation_interval: 15s
+    |scrape_configs:
+    |- job_name: templeuser
+    |  static_configs:
+    |  - targets:
+    |    - templeuser:1025
+    |""".stripMargin
 }
--- a/src/main/scala/temple/builder/project/ProjectBuilder.scala
+++ b/src/main/scala/temple/builder/project/ProjectBuilder.scala
@@ -13,6 +13,8 @@ import temple.generate.docker.DockerfileGenerator
 import temple.generate.kube.KubernetesGenerator
 import temple.generate.metrics.grafana.ast.Datasource
 import temple.generate.metrics.grafana.{GrafanaDashboardConfigGenerator, GrafanaDashboardGenerator, GrafanaDatasourceConfigGenerator}
+import temple.generate.metrics.prometheus.PrometheusConfigGenerator
+import temple.generate.metrics.prometheus.ast.PrometheusJob
 import temple.utils.StringUtils
 
 object ProjectBuilder {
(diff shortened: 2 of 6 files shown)


src/e2e/scala/temple/SimpleE2ETestData.scala
```diff
@@ -245,4 +245,15 @@ object SimpleE2ETestData {
       |  isDefault: true
       |  editable: true
       |""".stripMargin
+
+  val prometheusConfig: String =
+    """global:
```
c-harding, line 280: This continues to baffle me, in my IntelliJ the `|` is aligned with the third `"`

src/main/scala/temple/builder/project/ProjectBuilder.scala
```diff
@@ -54,19 +56,20 @@ object ProjectBuilder {
...
-        val dockerfileRoot     = DockerfileBuilder.createServiceDockerfile(name.toLowerCase, service, port)
+        val dockerfileRoot     = DockerfileBuilder.createServiceDockerfile(name.toLowerCase, service, port.service)
         val dockerfileContents = DockerfileGenerator.generate(dockerfileRoot)
         (File(s"${name.toLowerCase}", "Dockerfile"), dockerfileContents)
     }
 
     val orchestrationRoot = OrchestrationBuilder.createServiceOrchestrationRoot(
       StringUtils.kebabCase(templefile.projectName),
-      templefile.servicesWithPorts.toSeq,
+      templefile.servicesWithPorts.map { case (name, block, ports) => (name, block, ports.service) }.toSeq,
```
c-harding, line 66: Not a fan of this, but not sure how it could be made nicer. Maybe wrap it up into `templefile.servicesWithServicePorts`?

src/test/scala/temple/builder/project/ProjectBuilderTestData.scala
```diff
@@ -554,4 +565,15 @@ object ProjectBuilderTestData {
...
+  val complexTemplefilePrometheusConfig: String =
+    """global:
+    |  scrape_interval: 15s
+    |  evaluation_interval: 15s
+    |scrape_configs:
+    |- job_name: complexuser
+    |  static_configs:
+    |  - targets:
+    |    - complexuser:1025
+    |""".stripMargin
```
c-harding: It’s properly aligned on the addition at [l.309](https://github.com/TempleEight/temple/pull/180/files#diff-2db252505436751f5fb19b1bad244c85R309), but not here?

c-harding (approved):

src/main/scala/temple/builder/project/ProjectBuilder.scala
jaylees14: This was somewhat intentional - Kube will eventually need to know about the metrics ports too so it should be able to take the Ports type instead of Int. Could do that in a follow up?

src/test/scala/temple/builder/project/ProjectBuilderTestData.scala
jaylees14: Wish the formatter would sort this 😭

src/main/scala/temple/builder/project/ProjectBuilder.scala
c-harding: ok, fair enough. I wonder if it’s worth making a case class for the `(name, block, ports)` triple

src/test/scala/temple/builder/project/ProjectBuilderTestData.scala
Lewiky: Will this be fixed by #181 ?

Lewiky (approved):
Looks good bro - Think the Ports look good and probably necessary, though wish there were a nicer way to just get the one port without doing (name, block, ports) => (name, block, ports.foo)

PR #183, opened 2020-03-27
[cmd] Take input from stdin

c-harding:
This allows you to pipe input in: `temple validate -`. As a result, any shell scripting with generating Templefiles on the fly should be slightly easier


jaylees14 (approved):
Me likey 👌

willjvsmith (approved):
Nice, LGTM :rocket:

PR #184, opened 2020-03-27
[endpoint-test] Add test CLI mode

jaylees14:
I'm going to do this one in reverse from how we would normally - starting with the CLI and then adding more features from there.

Not sure I like `-g` for `--generated`, thoughts?

### DIFF
--- a/src/main/scala/temple/Main.scala
+++ b/src/main/scala/temple/Main.scala
@@ -16,6 +16,7 @@ object Main extends App {
     val config = new TempleConfig(args)
     config.subcommand match {
       case Some(config.Generate) => Application.generate(config)
+      case Some(config.Test)     => Application.test(config)
       case Some(_)               => throw new TempleConfig.UnhandledArgumentException
       case None                  => config.printHelp()
     }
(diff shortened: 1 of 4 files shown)


src/main/scala/temple/test/EndpointTester.scala
```diff
@@ -0,0 +1,9 @@
+package temple.test
+
+import temple.ast.Templefile
+
+object EndpointTester {
+
+  def test(templefile: Templefile, generatedPath: String): Unit =
+    println("Testing templefile...")
```
c-harding, line 8: Best test

src/main/scala/temple/Main.scala
```diff
@@ -16,6 +16,7 @@ object Main extends App {
     val config = new TempleConfig(args)
     config.subcommand match {
       case Some(config.Generate) => Application.generate(config)
+      case Some(config.Test)     => Application.test(config)
```
c-harding, line 18: I predict a conflict (plz merge dev in)

Lewiky (approved):
LGTM - could use `-d` and `--dir` for the generated directory?

c-harding (approved):

src/main/scala/temple/test/EndpointTester.scala
willjvsmith: The best test

willjvsmith (approved):

PR #185, opened 2020-03-27
[go-gen] Generate auth dao.go

willjvsmith:
* Adds generation of dao.go for the auth service

To be honest this felt like a waste of time, as we are not going to add support for other authTypes (username etc.), and we're not going to change from using UUID for IDs. Therefore this could have been read from file 🤷‍♂


src/main/scala/temple/generate/server/go/auth/GoAuthServiceDAOGenerator.scala
```diff
@@ -1,9 +1,189 @@
...
+      GoCommonDAOGenerator.generateExecuteQuery(),
+    )
+
+  private def generateCreateFunction(root: AuthServiceRoot, scanFunctionCall: String): String = {
+    val queryStatement = genDeclareAndAssign(
+      genFunctionCall(
+        "executeQueryWithRowResponse",
+        "dao.DB",
+        doubleQuote(
+          s"INSERT INTO auth (${root.idAttribute.name}, ${root.authAttribute.authType.name}, password) VALUES ($$1, $$2, $$3) RETURNING id, name, password",
```
jaylees14: Always gonna hardcode this, rather than switching on database? I guess applies for imports too

jaylees14 (approved):
Changes look good. Only comment is on the hardcoding of SQL queries. I know the likelihood of us adding anything else is virtually zero, but just wondered if you had any plans?

src/main/scala/temple/generate/server/go/auth/GoAuthServiceDAOGenerator.scala
```diff
@@ -1,9 +1,189 @@
...
+        "",
+        "// pq acts as the driver for SQL requests",
+        doubleQuote("github.com/lib/pq"),
+      ),
+    )
+
+  private[auth] def generateGlobals(): String =
+    mkCode.lines(
+      "// https://www.postgresql.org/docs/9.3/errcodes-appendix.html",
+      genConst("psqlUniqueViolation", doubleQuote("unique_violation")),
```
c-harding, line 31: Magic strings 😍

src/main/scala/temple/generate/server/go/auth/GoAuthServiceDAOGenerator.scala
c-harding: And are we no longer switching on interpolation mode either?

willjvsmith: I'll move the the SQL queries to be taken in by the `AuthServiceRoot`

willjvsmith: It's hard to know what will be common between different datastores, so I thought rather than guess I would hardcode it

src/main/scala/temple/generate/server/go/auth/GoAuthServiceDAOGenerator.scala
```diff
@@ -1,9 +1,189 @@
...
+
+  private def generateCreateFunction(root: AuthServiceRoot, scanFunctionCall: String): String = {
+    val queryStatement = genDeclareAndAssign(
+      genFunctionCall(
+        "executeQueryWithRowResponse",
+        "dao.DB",
+        doubleQuote(
+          s"INSERT INTO auth (${root.idAttribute.name}, ${root.authAttribute.authType.name}, password) VALUES ($$1, $$2, $$3) RETURNING id, name, password",
+        ),
+        s"input.${root.idAttribute.name.toUpperCase()}",
```
c-harding:
```suggestion
        s"input.${root.idAttribute.name.toUpperCase}",
```
feels weirdly inconsistent with the line below

src/main/scala/temple/generate/server/go/auth/GoAuthServiceDAOGenerator.scala
```diff
@@ -1,9 +1,189 @@
...
+            "switch err",
+            CodeWrap.curly.noIndent(
+              "case sql.ErrNoRows:",
+              tabIndent(genReturn("nil", "ErrAuthNotFound")),
+              "default:",
+              tabIndent(genReturn("nil", "err")),
+            ),
+          ),
+        ),
+      ),
```
c-harding: Are we not factoring this out then?

c-harding (approved):
All looks good to me

src/main/scala/temple/generate/server/go/auth/GoAuthServiceDAOGenerator.scala
willjvsmith: just for u

PR #187, opened 2020-03-27
[go-gen] Add non-auth handler.go generation

willjvsmith:
* Adds generation of handler.go for non-auth services

Once this is in, I think I've only got the main functions and the handlers left

### DIFF
--- a/src/main/scala/temple/generate/server/go/service/GoServiceCommGenerator.scala
+++ b/src/main/scala/temple/generate/server/go/service/GoServiceCommGenerator.scala
@@ -17,7 +17,7 @@ object GoServiceCommGenerator {
         doubleQuote("fmt"),
         doubleQuote("net/http"),
         "",
-        s""""${root.module}/util"""",
+        doubleQuote(s"${root.module}/util"),
         doubleQuote("github.com/google/uuid"),
       ),
     )
--- a/src/main/scala/temple/generate/server/go/service/GoServiceCommGenerator.scala
+++ b/src/main/scala/temple/generate/server/go/service/GoServiceCommGenerator.scala
@@ -1,19 +1,91 @@
 package temple.generate.server.go.service
 
 import temple.generate.server.ServiceRoot
+import temple.generate.server.go.common.GoCommonGenerator._
 import temple.generate.utils.CodeTerm.{CodeWrap, mkCode}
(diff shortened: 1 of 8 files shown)


src/main/scala/temple/generate/server/go/service/GoServiceCommGenerator.scala
```diff
@@ -1,19 +1,91 @@
...
-  )
+  private[service] def generateImports(root: ServiceRoot): String =
+    mkCode(
+      "import",
+      CodeWrap.parens.tabbed(
+        doubleQuote("errors"),
+        doubleQuote("fmt"),
+        doubleQuote("net/http"),
+        "",
+        s""""${root.module}/util"""",
```
jaylees14: Am I missing something, but can't this be `doubleQuote(s"${root.module}/util")`?

src/main/scala/temple/generate/server/go/service/GoServiceCommGenerator.scala
```diff
@@ -1,19 +1,91 @@
...
+    )
+
+  private[service] def generateCommFunctionDecl(root: ServiceRoot, serviceName: String): String =
+    mkCode(
+      genFunctionCall(
+        s"Check${serviceName.capitalize}",
+        s"${serviceName}${root.idAttribute.name.toUpperCase} ${generateGoType(root.idAttribute.attributeType)}",
+        "token string",
+      ),
+      "(bool, error)",
```
jaylees14, line 32: I wonder if some argument labels would be useful here? Would make it easier to read. Haven't done it elsewhere tho so I guess would be a pain

jaylees14 (approved):
LGTM, just a couple of pedantic comments

c-harding (approved):

src/main/scala/temple/generate/server/go/service/GoServiceCommGenerator.scala
willjvsmith: whoops, that was the only import line I left in from the old code

src/main/scala/temple/generate/server/go/service/GoServiceCommGenerator.scala
willjvsmith: IntelliJ sticks them in, but not so readable on GitHub

PR #188, opened 2020-03-27
[endpoint-test] Spin up Kube infra, perform sample request and shut down

jaylees14:
This spins up the Kube platform, makes a sample request (more to come later) and shuts down again.

Currently have this working with `spec-golang` and the following command:

```
temple test -d path/to/spec-golang path/to/example-templefiles/tinder.temple
```

Several assumptions are made, but will be relaxed in future PRs:
- They're deploying on Kubernetes and 
    - The user has `minikube` and `kubectl` preinstalled
- The generated code has a service called `user` 

This obviously won't work with any other service generated... because there's no code to run... Hopefully it will do in the future tho 🤭

I couldn't work out why running `source $generatedPath/deploy.sh` wouldn't correctly call the `configure-kong` script correctly - something to do with environment variables not being shared between them. I'm thinking of rebuilding the deploy script to be somewhat more robust to these changes, as well as giving some more status about the Kubernetes deployment status (for example at the moment it just hangs if something goes wrong) but for now this works.


src/main/scala/temple/test/EndpointTester.scala
```diff
@@ -2,8 +2,57 @@ package temple.test
...
+import sys.process._
+import scalaj.http.Http
+
 object EndpointTester {
 
+  class EnvironmentVariableNotSetException(name: String)
+      extends RuntimeException(s"The environment variable $name was required, but not found")
+
+  private def exec(command: String): String =
+    s"sh -c '$command'".!!
```
c-harding: what is Scala?

src/main/scala/temple/test/EndpointTester.scala
```diff
@@ -2,8 +2,57 @@ package temple.test
...
+
+  private def performSetup(templefile: Templefile, generatedPath: String): EndpointConfig = {
+    println("🍪 Spinning up Kubernetes infrastructure...")
+    // Require certain env vars to be set
+    Seq("REG_URL", "REG_USERNAME", "REG_PASSWORD", "REG_EMAIL").foreach { envVar =>
+      if (Option(System.getenv(envVar)).isEmpty) throw new EnvironmentVariableNotSetException(envVar)
```
c-harding: is this really preferable to `== null`?

src/main/scala/temple/test/EndpointTester.scala
```diff
@@ -2,8 +2,57 @@ package temple.test
...
+    println(sampleRequest)
+  }
+
   def test(templefile: Templefile, generatedPath: String): Unit =
-    println("Testing templefile...")
+    try {
+      val config = performSetup(templefile, generatedPath)
+      performTests(templefile, generatedPath, config.baseIP)
+    } finally {
+      performShutdown(templefile, generatedPath)
```
c-harding: Is running this going to be problematic if `performSetup` dies over an env var, so nothing’s started?

c-harding (approved):

src/main/scala/temple/test/EndpointTester.scala
jaylees14: https://www.scala-lang.org/api/current/scala/sys/process/ProcessBuilder.html Behold...

src/main/scala/temple/test/EndpointTester.scala
jaylees14: I guess not, I'll update

src/main/scala/temple/test/EndpointTester.scala
jaylees14: The command will fail, which won't cause anything bad to happen but might produce a weird error message, so I'll move the env var check out

jaylees14:
Also... I'm tempted to blacklist this file from code coverage, since it's probably has too many side effects to be tested. Thoughts @xsanda ?

src/main/scala/temple/test/ProjectTester.scala
```diff
@@ -0,0 +1,70 @@
...
+import temple.test.internal.EndpointConfig
+
+object ProjectTester {
+
+  class EnvironmentVariableNotSetException(name: String)
+      extends RuntimeException(s"The environment variable $name was required, but not found")
+
+  /** Execute a command, returning the stdout */
+  private def exec(command: String): String =
+    s"sh -c '$command'".!!
```
willjvsmith, line 16: interesting syntax

willjvsmith (approved):
LGTM with regard to the spec you outlined in the PR description.

Just to check I'm on the same page, this is for checking temple file against what is generated right

PR #189, opened 2020-03-28
[project-test] Add Auth Service Tests

jaylees14:
If a project has an auth service with email, test the `register` and `auth` endpoints.

Test plan:
```
$ temple test -d ../spec-golang ../example-templefiles/tinder.temple
...
🍪 Spinning up Kubernetes infrastructure...
🧪 Testing auth service
    ✅ auth register
    ✅ auth login
💀 Shutting down Kubernetes infrastructure...
```

### DIFF
--- a/src/main/scala/temple/test/ProjectTester.scala
+++ b/src/main/scala/temple/test/ProjectTester.scala
@@ -22,9 +22,10 @@ object ProjectTester {
     exec(s"sh $generatedPath/deploy.sh")
 
     // Grab Kong's URL for future requests
-    val urls   = exec("minikube service kong --url").split("\n")
-    val baseIP = urls(0).replaceAll("http[s]*://", "")
-    val config = ProjectConfig(baseIP, urls(1))
+    val minikubeURLString = exec("minikube service kong --url")
+    val urls              = minikubeURLString.split("\n")
+    val baseIP            = urls(0).replaceAll("http[s]*://", "")
+    val config            = ProjectConfig(baseIP, urls(1))
 
     // Running deploy.sh doesn't set Kong up correctly for some reason. I couldn't figure it out, so manually call that part
     exec(
--- a/src/main/scala/temple/test/ProjectTester.scala
+++ b/src/main/scala/temple/test/ProjectTester.scala
@@ -22,10 +22,9 @@ object ProjectTester {
     exec(s"sh $generatedPath/deploy.sh")
 
     // Grab Kong's URL for future requests
-    val minikubeURLString = exec("minikube service kong --url")
-    val urls              = minikubeURLString.split("\n")
-    val baseIP            = urls(0).replaceAll("http[s]*://", "")
-    val config            = ProjectConfig(baseIP, urls(1))
+    val Array(baseURL, kongAdmin, _*) = exec("minikube service kong --url").split("\n")
+    val baseIP                        = baseURL.replaceAll("http[s]*://", "")
+    val config                        = ProjectConfig(baseIP, kongAdmin)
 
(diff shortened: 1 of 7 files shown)


src/main/scala/temple/test/ProjectTester.scala
```diff
@@ -16,15 +16,15 @@ object ProjectTester {
...
 
   /** Configure the infrastructure for testing */
-  private def performSetup(templefile: Templefile, generatedPath: String): EndpointConfig = {
+  private def performSetup(templefile: Templefile, generatedPath: String): ProjectConfig = {
     println("🍪 Spinning up Kubernetes infrastructure...")
     // Deploy Kubernetes
     exec(s"sh $generatedPath/deploy.sh")
 
     // Grab Kong's URL for future requests
     val urls   = exec("minikube service kong --url").split("\n")
```
c-harding: Can we destructure this into variables?

src/main/scala/temple/test/internal/EndpointTest.scala
```diff
@@ -0,0 +1,18 @@
...
+
+  def assertEqual(expected: Any, found: Any): Unit =
+    if (!expected.equals(found))
+      fail(s"Expected $expected but found $found")
+
+  def assert(value: Boolean, failMsg: String): Unit =
+    if (!value) fail(failMsg)
+
+  def pass(): Unit =
+    println(s"\t✅ $service $endpointName")
```
c-harding: `\t` will be inconsistently wide between terminals, does this look very weird if a tab is expanded to width eight?

src/main/scala/temple/test/internal/ServiceTestUtils.scala
```diff
@@ -0,0 +1,40 @@
...
+    * @param url The URL to send the request to, prefixed with the protocol
+    * @param body The JSON body to send along with the request
+    * @param test The endpoint under test
+    * @return The decoded JSON response from the request
+    */
+  def postRequest(url: String, body: Json, test: EndpointTest): JsonObject = {
+    val response = Http(url).postData(body.toString).method("POST").asString
+    test.assertEqual(200, response.code)
+
+    parse(response.body) match {
```
c-harding: Would this block look better with fromEither and getOrElse, to avoid the nesting and to make it clearer where the return value is?

c-harding (approved):
Looks good, but can we not piggyback on an existing testing framework?

src/main/scala/temple/test/ProjectTester.scala
jaylees14: How do you mean? One for the exec and one for the split?

src/main/scala/temple/test/internal/EndpointTest.scala
jaylees14: If only we had an indent function... 😉

src/main/scala/temple/test/internal/ServiceTestUtils.scala
```diff
@@ -0,0 +1,41 @@
...
+  /**
+    * Execute a POST request, where the response body is expected to be a JSON object
+    * @param url The URL to send the request to, prefixed with the protocol
+    * @param body The JSON body to send along with the request
+    * @param test The endpoint under test
+    * @return The decoded JSON response from the request
+    */
+  def postRequest(url: String, body: Json, test: EndpointTest): JsonObject = {
+    val response = Http(url).postData(body.toString).method("POST").asString
+    test.assertEqual(200, response.code)
```
willjvsmith, line 20: Probs replace status codes with enum for when they get more exotic

willjvsmith (approved):
LGTM 🚀 

Guess we're succumbing to only supporting email 😉 makes sense with our time frame

PR #193, opened 2020-03-30
[Go-Map] Server Builder

Lewiky:
2nd go-map PR - implements the ServerBuilder object to construct ServiceBlocks to pass to the go generator - doesn't currently get called from the ProjectBuilder as this will come in the third PR.

### DIFF
--- a/src/main/scala/temple/builder/ServerBuilder.scala
+++ b/src/main/scala/temple/builder/ServerBuilder.scala
@@ -1,0 +1,74 @@
+package temple.builder
+
+import temple.ast.AttributeType.ForeignKey
+import temple.ast.Metadata.{ServiceEnumerable, ServiceLanguage}
+import temple.ast.{Attribute, AttributeType, Metadata, ServiceBlock}
+import temple.builder.project.LanguageConfig.GoLanguageConfig
+import temple.builder.project.ProjectConfig
+import temple.detail.LanguageDetail
+import temple.detail.LanguageDetail.GoLanguageDetail
+import temple.generate.CRUD._
+import temple.generate.database.{PostgresContext, PostgresGenerator, PreparedType}
+import temple.generate.server.{CreatedByAttribute, IDAttribute, ServiceRoot}
+
+import scala.collection.immutable.ListMap
+
+object ServerBuilder {
+
+  def buildServiceRoot(
+    serviceName: String,
+    serviceBlock: ServiceBlock,
+    port: Int,
+    endpoints: Set[CRUD],
+    detail: LanguageDetail,
+  ): ServiceRoot = {
+    val language = serviceBlock.lookupMetadata[ServiceLanguage].getOrElse(ProjectConfig.defaultLanguage)
+    val languageConfig = language match {
+      case ServiceLanguage.Go => GoLanguageConfig
+    }
+
+    val createdBy: CreatedByAttribute = serviceBlock.lookupMetadata[ServiceEnumerable] match {
+      case Some(ServiceEnumerable(true)) =>
+        CreatedByAttribute.EnumerateByCreator(
+          languageConfig.createdByInputName,
+          languageConfig.createdByName,
+          languageConfig.createdByAttributeType,
+        )
+      case Some(ServiceEnumerable(false)) =>
+        CreatedByAttribute.EnumerateByAll(
+          languageConfig.createdByInputName,
+          languageConfig.createdByName,
+          languageConfig.createdByAttributeType,
+        )
+      case None => CreatedByAttribute.None
+    }
+
+    val iDAttribute = IDAttribute("id", AttributeType.UUIDType)
+
+    implicit val context
+      : PostgresContext = PostgresContext(PreparedType.DollarNumbers) //TODO: Does this need to change with server lang?
+    val queries: ListMap[CRUD, String] =
+      DatabaseBuilder
+        .buildQuery(serviceName, serviceBlock, endpoints, iDAttribute, createdBy)
+        .map { case (crud, statement) => crud -> PostgresGenerator.generate(statement) }
+
+    val moduleName: String = detail match {
+      case GoLanguageDetail(modulePath) => s"$modulePath/$serviceName"
+    }
+    //TODO: Auth
+
+    ServiceRoot(
+      serviceName,
+      module = moduleName,
+      comms = serviceBlock.attributes.collect {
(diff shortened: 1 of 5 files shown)


src/main/scala/temple/builder/ServerBuilder.scala
```diff
@@ -0,0 +1,74 @@
...
+      case Some(ServiceEnumerable(false)) =>
+        CreatedByAttribute.EnumerateByAll(
+          languageConfig.createdByInputName,
+          languageConfig.createdByName,
+          languageConfig.createdByAttributeType,
+        )
+      case None => CreatedByAttribute.None
+    }
+
+    val iDAttribute = IDAttribute("id", AttributeType.UUIDType)
```
jaylees14:
```suggestion
    val idAttribute = IDAttribute("id", AttributeType.UUIDType)
```

src/main/scala/temple/builder/ServerBuilder.scala
```diff
@@ -0,0 +1,74 @@
...
+
+    implicit val context
```
jaylees14: Shouldn't we be switching on database?

src/main/scala/temple/builder/ServerBuilder.scala
```diff
@@ -0,0 +1,74 @@
...
+      : PostgresContext = PostgresContext(PreparedType.DollarNumbers) //TODO: Does this need to change with server lang?
+    val queries: ListMap[CRUD, String] =
+      DatabaseBuilder
+        .buildQuery(serviceName, serviceBlock, endpoints, iDAttribute, createdBy)
+        .map { case (crud, statement) => crud -> PostgresGenerator.generate(statement) }
+
+    val moduleName: String = detail match {
```
jaylees14, line 62: This also feels quite Go - should be inside a switch?

src/main/scala/temple/builder/ServerBuilder.scala
```diff
@@ -0,0 +1,74 @@
...
+      case GoLanguageDetail(modulePath) => s"$modulePath/$serviceName"
+    }
+    //TODO: Auth
+
+    ServiceRoot(
+      serviceName,
+      module = moduleName,
+      comms = serviceBlock.attributes.collect {
+        case (_, Attribute(x: ForeignKey, _, _)) => x.references
+      }.toSeq,
```
jaylees14: I'd be tempted to pull this into a variable with a comment

src/main/scala/temple/builder/project/LanguageConfig.scala
```diff
@@ -0,0 +1,20 @@
...
+  val createdByInputName: String
+  val createdByName: String
+  val createdByAttributeType: AttributeType
+}
+
+object LanguageConfig {
+
+  /** The implicit values used to generate Golang code that don't come from the user */
+  object GoLanguageConfig extends LanguageConfig {
+    override val createdByInputName: String            = "authID"
```
jaylees14, line 20: What is this string actually representing? The database column name? The go struct name?

jaylees14 (commented):
Looks good - could we have some tests for the `createdBy` stuff?

src/main/scala/temple/builder/ServerBuilder.scala
Lewiky: Well the moduleName needs to be defined in every ServiceBlock anyway - so it needs to be set to something. And the detail matching checks for a GoDetail which should only happen when the server lang is go - though I could put an extra check here if you think that's needed

jaylees14: Ah yes I remember this now actually - we agreed to take all the parameters for every `Service`. No worries, leave it as is

src/main/scala/temple/builder/project/LanguageConfig.scala
Lewiky: I actually have no idea, @smithwjv explained it to me at one point but I didn't really understand - waiting for an explanation

Lewiky: Update: Got a response and have added a comment to the base `LanguageConfig` explaining each param

jaylees14 (approved):
LGTM, thanks for making those tweaks :)

src/test/scala/temple/builder/ServerBuilderTest.scala
```diff
@@ -0,0 +1,135 @@
...
+            detail = GoLanguageDetail("github.com/squat/and/dab"),
+          )
+    }
+    serviceRoot shouldBe ServiceRoot(
+      "test-service",
+      "github.com/squat/and/dab/test-service",
+      comms = Seq(),
+      port = 1024,
+      opQueries = ListMap(
+        Create -> "INSERT INTO test-service (id, bankBalance, name, isStudent, dateOfBirth, timeOfDay, expiry, image) VALUES ($1, $2, $3, $4, $5, $6, $7, $8) RETURNING id, bankBalance, name, isStudent, dateOfBirth, timeOfDay, expiry, image;",
```
willjvsmith, line 35: Now that the SQL-Gen has been updated to handle all of these, we should probs change these (and the other test SQL strings) to be generated through it, rather than using strings.

src/main/scala/temple/builder/project/LanguageConfig.scala
```diff
@@ -0,0 +1,24 @@
+package temple.builder.project
+
+import temple.ast.AttributeType
+
+/** Defines the values used for generation in a language that do not come from the user
+  * inputName the name of the variable used as input, e.g. "authID", present for improving semantics of code
+  * name the name of attribute, e.g. "createdBy"
+  * attributeType type of the attribute, e.g. UUID
+  * */
+sealed trait LangnuageConfig {
```
willjvsmith: Langnuage

willjvsmith (approved):
LGTM 🚀 

Few minor things

src/main/scala/temple/builder/project/LanguageConfig.scala
Lewiky: This was the typo tehe

src/test/scala/temple/builder/ServerBuilderTest.scala
Lewiky: I'd argue that calls out to the SQL code too much though and loses the atomicity of the unit tests and starts feeling more integration testy

PR #194, opened 2020-03-30
Service tester crud

jaylees14:
Test the `CREATE` endpoint of each service by mocking a request body, then validating the response is what we expect:

Again running with `spec-golang` and `tinder.temple`
```
🍪 Spinning up Kubernetes infrastructure...
🧪 Testing Auth service
    ✅ Auth register
    ✅ Auth login
🧪 Testing User service
    ✅ User create
🧪 Testing Match service
    ✅ Match create
💀 Shutting down Kubernetes infrastructure...
```

### DIFF
--- a/src/main/scala/temple/test/internal/EndpointTest.scala
+++ b/src/main/scala/temple/test/internal/EndpointTest.scala
@@ -1,9 +1,97 @@
 package temple.test.internal
 
+import java.text.SimpleDateFormat
+import java.util.UUID
+
+import io.circe.{Json, JsonObject}
+import temple.ast.{Annotation, Attribute, AttributeType}
 import temple.utils.StringUtils
+
+import scala.util.Try
 
 private[internal] class EndpointTest(service: String, endpointName: String) {
   class TestFailedException(msg: String) extends RuntimeException(msg)
+
+  // Validate that the response JSON for the provided key matches the Attribute
+  private def validateResponseType(key: String, responseForKey: Json, attribute: Attribute): Unit =
+    attribute.attributeType match {
+      case AttributeType.ForeignKey(_) | AttributeType.UUIDType =>
+        val stringValue = responseForKey.asString.getOrElse(fail(s"$key is not a valid string"))
+        assert(Try(UUID.fromString(stringValue)).isSuccess, s"$key is not a valid UUID")
+      case AttributeType.BoolType =>
+        assert(responseForKey.isBoolean, s"$key is not of type bool")
+      case AttributeType.DateType =>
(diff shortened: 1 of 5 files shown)


src/main/scala/temple/test/internal/EndpointTest.scala
```diff
@@ -1,10 +1,98 @@
...
+        max.foreach { maxValue =>
+          assert(floatValue <= maxValue, s"$key does not satisfy max bound of $maxValue")
+        }
+    }
+
+  def validateResponseBody(request: JsonObject, response: JsonObject, attributes: Map[String, Attribute]): Unit = {
+    // The response should contain exactly the keys in attributes, PLUS an ID attribute, MINUS anything that is @server
+    val expectedAttributes = attributes.filter {
+      case (_, attribute) =>
+        attribute.accessAnnotation.fold(true) {
```
c-harding: One of these would surely be preferable to explicitly listing every option:

accessAnnotation != Some(Annotation.Server)
!accessAnnotation.contains(Annotation.Server)

src/main/scala/temple/test/internal/EndpointTest.scala
```diff
@@ -1,10 +1,98 @@
...
+          case Annotation.Server                        => false
+          case Annotation.ServerSet | Annotation.Client => true
+        }
+    }
+    val expectedAttributeNames = (expectedAttributes.keys ++ Seq("ID")).map(_.capitalize).toSet
```
c-harding: Flip the args of the ++, replace Seq with Set, and there’s no need for the toSet.

src/main/scala/temple/test/internal/EndpointTest.scala
```diff
@@ -1,10 +1,98 @@
...
+    assertEqual(expectedAttributeNames, foundAttributeNames)
+
+    expectedAttributes.foreach {
+      case (name, attribute) =>
+        // Validate that the response JSON is of the expected types
+        val responseForKey = response(name.capitalize).getOrElse(fail(s"could not find key $name in response body"))
+        validateResponseType(name, responseForKey, attribute)
+
+        // Where the attribute is not @serverSet, validate that what was sent is exactly what is returned
+        val shouldValidateRequest = attribute.accessAnnotation.fold(true) {
```
c-harding: Ditto

c-harding (approved):

src/main/scala/temple/test/internal/EndpointTest.scala
jaylees14: Good shout

Lewiky (approved):
Little hard to follow in places, but overall looks good

PR #196, opened 2020-03-30
[Go-map] Update Project Builder & Tests 

Lewiky:
Calls the ServerBuilder from the main Application entrypoint and provides end to end and unit tests for the whole thing 

Currently three copies of all the go gen files here in the forms of:
* simple-temple projectbuilder unit tests
* complex-temple projectbuilder unit tests
* end to end tests

If anyone can give advice on how to break this up without breaking CI would be welcome

### DIFF
--- a/src/main/scala/temple/Main.scala
+++ b/src/main/scala/temple/Main.scala
@@ -2,6 +2,8 @@ package temple
 
 import java.nio.file.FileAlreadyExistsException
 
+import temple.detail.PoliceSergeantNicholasAngel
+
 /** Main entry point into the application */
 object Main extends App {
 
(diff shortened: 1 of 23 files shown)


src/main/scala/temple/Main.scala
```diff
@@ -13,7 +15,7 @@ object Main extends App {
   try {
     val config = new TempleConfig(args)
     config.subcommand match {
-      case Some(config.Generate) => Application.generate(config)
+      case Some(config.Generate) => Application.generate(config, PoliceSergeantNicholasAngel)
```
c-harding, line 18: Wtf is this and why is it the default behaviour of our program?

Lewiky: Mornin angle

PR #198, opened 2020-03-30
[go-gen] Remove attribute type from IDAttribute and CreatedByAttribute

willjvsmith:
* Removes attributeType field from IDAttribute and CreatedByAttribute, hardcoding the type to UUID

Means I can assume the type to be UUID in my generation, making it safer.


jaylees14 (approved):

src/main/scala/temple/generate/server/ServiceRoot.scala
```diff
@@ -14,7 +14,7 @@ import scala.collection.immutable.ListMap
   * @param comms a sequence of service names this service communicates with
   * @param opQueries a map of CRUD operations to their corresponding datastore query
   * @param port the port number this service will be served on
-  * @param idAttribute the name and type of the ID field in this service // TODO: This is assumed to be the same for all services (for comms), needs to be refactored
```
c-harding, line 17: and thus it was done

c-harding (approved):
lovely

PR #199, opened 2020-03-30
[Go-Map] Update ProjectBuilder & Tests 2: Electric Boogalloo

Lewiky:
Same as before, except this time all of the e2e tests have been removed to help minimise the diff


src/test/resources/go/complex-user/dao/dao.go.snippet
```diff
@@ -0,0 +1,172 @@
...
+	"github.com/squat/and/dab/complexuser/util"
+	"github.com/google/uuid"
+
+	// pq acts as the driver for SQL requests
+	_ "github.com/lib/pq"
+)
+
+// Datastore provides the interface adopted by the DAO, allowing for mocking
+type Datastore interface {
+	CreateComplexuser(input CreateComplexuserInput) (*Complexuser, error)
```
jaylees14, line 17: Don't think this is you, probably @smithwjv - could we make this `ComplexUser`?

jaylees14 (approved):
All looks good to me - obviously can't run the Go files, but a quick read and everything looks sensible.

Do wish we supported `SELECT * FROM table` tho 😝

src/test/resources/go/complex-user/dao/dao.go.snippet
willjvsmith: Should be handled if it comes in as `complexUser`

willjvsmith: It capitalises whatever comes in for the service name

c-harding (changes requested):
*Boogaloo

src/test/resources/go/complex-user/dao/dao.go.snippet
c-harding: This is a coding error, nowhere in the input does it appear with a lowercase `u`.

src/main/scala/temple/Main.scala
```diff
@@ -13,7 +15,7 @@ object Main extends App {
   try {
     val config = new TempleConfig(args)
     config.subcommand match {
-      case Some(config.Generate) => Application.generate(config)
+      case Some(config.Generate) => Application.generate(config, PoliceSergeantNicholasAngel)
```
c-harding, line 18: I still think this should be a mock and not the main application

c-harding (approved):

src/main/scala/temple/Main.scala
Lewiky: That was gonna come in a separate PR

PR #202, opened 2020-03-31
[fix] Map foreign keys to uuid types in server builder

Lewiky:
I forgot that the server code requires the ForeignKey attributes to be turned into UUIDs at the builder stage - This fixes that

n.b we should probably update the tests to include some templefiles with foreign keys (though this sounds like a whole lot of work)


jaylees14 (approved):
LGTM, I agree with your extending the tests remarks

willjvsmith (approved):
LGTM 🚀 

I probably should have used the ForeignKey type when generating, but atm there's no distinction, as UUIDs in the attributes argument are always foreign keys

PR #203, opened 2020-03-31
[service-test] Remove server and server set items from request body

jaylees14:
Filter attributes so that server and server set attributes are not sent in a request body


src/main/scala/temple/test/internal/ServiceTestUtils.scala
```diff
@@ -110,42 +110,51 @@ object ServiceTestUtils {
...
-        }
-        (name, value)
-    }.asJson
+    // Drop any attributes that are @server or @serverSet
+    attributes
+      .filterNot {
+        case (_, attribute) =>
+          attribute.accessAnnotation.contains(Annotation.Server) || attribute.accessAnnotation.contains(
+            Annotation.ServerSet,
+          )
```
c-harding, line 119: are you able to break this after the `||` instead?

c-harding (approved):
nice :)

Lewiky (approved):

src/main/scala/temple/test/internal/ServiceTestUtils.scala
jaylees14: Actually no, for some reason the compiler doesn't like it:

```
Error:(118, 23) ';' expected but '.' found.
          || attribute.accessAnnotation.contains(Annotation.ServerSet)
```

PR #204, opened 2020-03-31
[go-gen] Add main.go request struct generation

willjvsmith:
* Adds generation of request struct generation in the non-auth service main files
  * Adds placeholder parameter validation, ready for @jaylees14 
  * Adds temporary checks in `generateImports` to make the integration tests pass
* Updates unit, e2e and integration tests

This stuff gets quite tricky, but I think once these structs are done it should get a bit easier

### DIFF
--- a/src/main/scala/temple/generate/server/go/service/GoServiceMainGenerator.scala
+++ b/src/main/scala/temple/generate/server/go/service/GoServiceMainGenerator.scala
@@ -1,8 +1,9 @@
 package temple.generate.server.go.service
 
+import temple.ast.{Annotation, Attribute, AttributeType}
+import temple.generate.CRUD
 import temple.generate.CRUD.CRUD
 import temple.generate.server.ServiceRoot
-import temple.generate.utils.CodeTerm
 import temple.generate.utils.CodeTerm.{CodeWrap, mkCode}
 import temple.utils.StringUtils.doubleQuote
 import temple.generate.server.go.common.GoCommonGenerator._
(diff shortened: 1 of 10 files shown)


src/main/scala/temple/generate/server/go/service/GoServiceMainGenerator.scala
```diff
@@ -21,12 +27,28 @@ object GoServiceMainGenerator {
...
         //doubleQuote("net/http"),
+
+        // TODO: This check is temporary to make the integration tests pass
+        when(
+          clientAttributes
+            .exists {
+              case (_, attr) =>
+                attr.attributeType == AttributeType.DateType ||
+                attr.attributeType == AttributeType.TimeType ||
+                attr.attributeType == AttributeType.DateTimeType
```
c-harding, line 38: personally it feels like `Seq(AttributeType.DateType,AttributeType.TimeType,AttributeType.DateTimeType) contains attr.attributeType` would be nicer, but happy to be overruled

jaylees14 (approved):
LGTM 👍

src/main/scala/temple/generate/server/go/service/GoServiceMainGenerator.scala
```diff
@@ -21,12 +27,28 @@ object GoServiceMainGenerator {
         //doubleQuote("fmt"),
         //doubleQuote("log"),
         //doubleQuote("net/http"),
+
+        // TODO: This check is temporary to make the integration tests pass
+        when(
+          clientAttributes
+            .exists {
+              case (_, attr) =>
```
c-harding, line 35: Likewise `clientAttributes.valuesIterator.exists { attr =>` feels nicer to me

c-harding (approved):
Looks nice!

src/main/scala/temple/generate/server/go/service/GoServiceMainGenerator.scala
willjvsmith: temporary for integration tests to pass, see TODO comment in the lines above

PR #205, opened 2020-03-31
[Validator] Test sets of errors rather than strings

c-harding:
The test messages were annoying me, so I’ve switched to testing the sets of error strings rather than the resulting error message. Don’t worry, the old method is still tested as before

Turn off whitespace to shrink the diff to a more manageable +90–69

### DIFF
--- a/src/test/scala/temple/DSL/semantics/ValidatorTest.scala
+++ b/src/test/scala/temple/DSL/semantics/ValidatorTest.scala
@@ -1,7 +1,7 @@
 package temple.DSL.semantics
 
 import org.scalatest.{FlatSpec, Matchers}
-import temple.DSL.semantics.Validator.validate
+import temple.DSL.semantics.Validator._
 import temple.DSL.semantics.ValidatorTest._
 import temple.ast.Annotation
 import temple.ast.Annotation.{Nullable, Unique}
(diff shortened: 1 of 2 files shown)


src/test/scala/temple/DSL/semantics/ValidatorTest.scala
```diff
@@ -14,210 +14,230 @@ class ValidatorTest extends FlatSpec with Matchers {
...
-              metadata = Seq(Writable.All, Readable.This),
-            ),
-          ),
+        templefileWithUserAttributes(
+          "b" -> Attribute(FloatType(max = Some(0), min = Some(1))),
+          "c" -> Attribute(FloatType(precision = -1)),
         ),
       )
-    } should have message singleError("#writable(all) is not compatible with #readable(this) in User")
+    } should have message {
```
jaylees14, line 236: I will never stop being amazed at what you can do with this language

jaylees14 (approved):
Looks good :)

src/test/scala/temple/DSL/semantics/ValidatorTest.scala
Lewiky: scala best language

Lewiky (approved):
Sure

PR #206, opened 2020-03-31
[go-gen] Populate validation strings

jaylees14:
Populate go validation strings where the library supports the constraints.

Currently it does not support everything that `Templefile` supports - such as when an upper bound is provided but a lower bound is not. This can be rectified with a custom validator, something along the lines of:

```go
type Foo struct {
    example int `valid:"lessThan(5)"`
}
valid.ParamTagMap["lessThan"] = govalidator.ParamValidator(func(str string, params ...string) bool {
    bound := params[0]
    return strconv.Atoi(str) <= bound
})
```

However, I felt this went a bit above and beyond the scope here - so I've left them as TODOs. 

Update the unit, integration & e2e tests as necessary

Also worth pointing out: as of TempleEight/spec-golang#101, byte arrays are sent as base64 encoded strings, the request types need to be updated here @smithwjv :)


Lewiky (approved):
LGTM boi - tho it's getting a little out of hand how many test file changes need to be made for every small go addition :S

c-harding:
Why does `ParamValidator` take an array of params?

src/e2e/resources/go/user/user.go.snippet
```diff
@@ -12,28 +12,28 @@ type env struct {
...
-	BirthDate          *time.Time `valid:"-"`
-	BreakfastTime      *time.Time `valid:"-"`
+	Username           *string    `valid:"type(string),required"`
+	Email              *string    `valid:"type(string),required,stringlength(5|40)"`
+	FirstName          *string    `valid:"type(string),required"`
+	LastName           *string    `valid:"type(string),required"`
+	CreatedAt          *time.Time `valid:"type(string),rfc3339WithoutZone,required"`
+	NumberOfDogs       *int32     `valid:"type(int32),required"`
+	CurrentBankBalance *float32   `valid:"type(float32),required"`
+	BirthDate          *time.Time `valid:"type(string),rfc3339WithoutZone,required"`
```
c-harding: Can `rfc3339WithoutZone` distinguish between date and datetime, and do we want to make that distinction?

src/main/scala/temple/generate/server/go/service/GoServiceMainGenerator.scala
```diff
@@ -64,16 +64,27 @@ object GoServiceMainGenerator {
...
-        case AttributeType.TimeType                       => "-"
-        case AttributeType.BlobType(size)                 => "-"
-        case AttributeType.StringType(max, min)           => "-"
-        case AttributeType.IntType(max, min, precision)   => "-"
-        case AttributeType.FloatType(max, min, precision) => "-"
+        case AttributeType.ForeignKey(_) | AttributeType.UUIDType => "type(string),uuid,required"
+        case AttributeType.BoolType                               => "type(bool),required"
+        case AttributeType.DateType                               => "type(string),rfc3339WithoutZone,required"
+        case AttributeType.DateTimeType                           => "type(string),rfc3339WithoutZone,required"
+        case AttributeType.TimeType                               => "type(string),rfc3339WithoutZone,required"
```
c-harding: These should surely be distinct, no? And for datetime, don’t we want a TZ?

c-harding (approved):
not sure about the date/time handling, otherwise good

PR #207, opened 2020-04-01
[go-gen] Add main.go response struct generation

willjvsmith:
* Adds generation of main.go (non-auth) structs
  * Moves struct gen into own file `GoServiceMainStructGenerator`
* Updates all tests

There's only about 60 lines of actual new code here (mainly`GoServiceMainStructGenerator:69` to EOF), the snippets take up a lot!

### DIFF
--- a/src/main/scala/temple/generate/server/go/service/main/GoServiceMainStructGenerator.scala
+++ b/src/main/scala/temple/generate/server/go/service/main/GoServiceMainStructGenerator.scala
@@ -1,54 +1,18 @@
-package temple.generate.server.go.service
+package temple.generate.server.go.service.main
 
+import temple.ast.AttributeType.{BlobType, DateTimeType, DateType, TimeType}
 import temple.ast.{Annotation, Attribute, AttributeType}
-import temple.ast.AttributeType._
 import temple.generate.CRUD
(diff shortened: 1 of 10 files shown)


src/test/scala/temple/testfiles/simple.temple
```diff
@@ -1,3 +1,4 @@
+
```
c-harding, line 1: ?

src/main/scala/temple/generate/server/go/service/main/GoServiceMainStructGenerator.scala
```diff
@@ -0,0 +1,114 @@
...
+    )
+
+  private def generateResponseStruct(root: ServiceRoot, operation: CRUD, fields: ListMap[String, String]): String =
+    mkCode.lines(
+      mkCode(
+        s"// ${operation.toString.toLowerCase}${root.name.capitalize}Response contains a",
+        operation match {
+          case Create => s"newly created ${root.name}"
+          case Read   => s"single ${root.name}"
+          case Update => s"newly updated ${root.name}"
```
jaylees14, line 94: What happens for a delete? I know we know there's never a request body, but feels like you should still deal with the case

c-harding (approved):
nice

jaylees14 (approved):
All looks good to me :)

PR #210, opened 2020-04-01
[Validator] Add decapitalize function

c-harding:
This is for adding the project name to the start of attributes with illegal names in the database. If any language doesn’t like a variable name, it gets renamed everywhere, so if Postgres doesn’t like a name, a prefix will be added, and this will still need to start with a lower-case name for Go


Lewiky (approved):
Noice

jaylees14 (approved):
What a surprise.. a regular expression 😉

PR #211, opened 2020-04-01
[Validator] Allow metadata to be looked up from struct blocks

c-harding:
This refactor of the struct blocks makes the validator a bit easier. It does come with the caveat that things like `#omit` will be inherited, but for those they shouldn’t be inherited from the project block either so `lookupLocalMetadata` can just be used instead for accessing this property for a struct block. We haven’t yet even implemented struct block handling though, so no big deal


Lewiky (approved):
coolio

jaylees14 (approved):

PR #212, opened 2020-04-01
[go-gen] Add main function generation

willjvsmith:
* Adds generation of the main function for both auth and non-auth services
  * Uses the same function for both since they're very similar, does mean deconstructing the root though
* Updates tests

This completes the `Auth` service! Should be ready to run aside from the `config.json`, which I'll do before the non-auth handlers.

### DIFF
--- a/src/main/scala/temple/generate/server/go/auth/GoAuthServiceGenerator.scala
+++ b/src/main/scala/temple/generate/server/go/auth/GoAuthServiceGenerator.scala
@@ -18,7 +18,7 @@ object GoAuthServiceGenerator extends AuthServiceGenerator {
         GoAuthServiceMainGenerator.generateImports(root),
         GoAuthServiceMainGenerator.generateStructs(),
         GoAuthServiceMainGenerator.generateRouter(),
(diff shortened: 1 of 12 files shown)


jaylees14 (approved):
LGTM 👍

src/main/scala/temple/generate/server/go/auth/GoAuthServiceGenerator.scala
```diff
@@ -19,7 +18,7 @@ object GoAuthServiceGenerator extends AuthServiceGenerator {
         GoAuthServiceMainGenerator.generateImports(root),
         GoAuthServiceMainGenerator.generateStructs(),
         GoAuthServiceMainGenerator.generateRouter(),
-        GoAuthServiceMainGenerator.generateMain(),
+        GoCommonMainGenerator.generateMain(isAuth = true, "auth", root.port, usesComms = true),
```
c-harding: it’s possible to name args, then subsequently not name future args? 😮

c-harding (approved):
nice, looks good

PR #214, opened 2020-04-01
[Go-Map] Auth Server Builder

Lewiky:
Updates the server builder to now also build AuthServiceRoot objects, prepared to generate the auth server files.


jaylees14 (approved):
Nice

c-harding (approved):
nice

PR #217, opened 2020-04-02
[go-gen][hooks] Add Hook and HookError structs

jaylees14:
Generate `Hook` struct, in addition to `HookError` and associated `Error` method.

Updates E2E, integration & unit tests.

### DIFF
--- a/src/it/scala/temple/containers/GolangSpec.scala
+++ b/src/it/scala/temple/containers/GolangSpec.scala
@@ -22,6 +22,7 @@ abstract class GolangSpec extends DockerShell2HttpService(8081) with DockerTestK
     val json = files.map { case (file, contents) => (file.folder + "/" + file.filename, contents) }.asJson.toString()
     Http(golangVerifyUrl)
       .params(Map("src" -> json, "root" -> entryFile.folder, "entrypoint" -> entryFile.filename))
+      .timeout(connTimeoutMs = 1000, readTimeoutMs = 10000)
       .asString
       .body
   }
--- a/src/main/scala/temple/generate/server/go/common/GoCommonGenerator.scala
+++ b/src/main/scala/temple/generate/server/go/common/GoCommonGenerator.scala
@@ -126,6 +126,23 @@ object GoCommonGenerator {
   private[go] def genMethodCall(objectName: String, methodName: String, args: String*): String =
     s"$objectName.${genFunctionCall(methodName, args: _*)}"
 
+  /** Generate method definition */
+  private[go] def genMethod(
+    objectName: String,
+    objectType: String,
+    methodName: String,
+    methodArgs: Seq[String],
+    methodReturn: Option[String],
+    methodBody: String,
+  ): String =
+    mkCode(
+      "func",
+      CodeWrap.parens(objectName, objectType),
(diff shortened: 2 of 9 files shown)


src/it/scala/temple/containers/GolangSpec.scala
```diff
@@ -22,6 +22,7 @@ abstract class GolangSpec extends DockerShell2HttpService(8081) with DockerTestK
     val json = files.map { case (file, contents) => (file.folder + "/" + file.filename, contents) }.asJson.toString()
     Http(golangVerifyUrl)
       .params(Map("src" -> json, "root" -> entryFile.folder, "entrypoint" -> entryFile.filename))
+      .timeout(connTimeoutMs = 1000, readTimeoutMs = 10000)
```
jaylees14, line 25: @xsanda - this should have fixed the error we were seeing with Travis last night. The time taken to compile the Go exceeded the read timeout of the HTTP request, so I've bumped the default read value to 10000ms

src/main/scala/temple/generate/server/go/common/GoCommonGenerator.scala
```diff
@@ -126,6 +126,23 @@ object GoCommonGenerator {
...
+    objectName: String,
+    objectType: String,
+    methodName: String,
+    methodArgs: Seq[String],
+    methodReturn: Option[String],
+    methodBody: String,
+  ): String =
+    mkCode(
+      "func",
+      CodeWrap.parens(objectName, objectType),
```
c-harding, line 140: Should objectType deal with the star here instead?

c-harding (approved):
nice

src/main/scala/temple/generate/server/go/common/GoCommonGenerator.scala
jaylees14: Not necessarily, since you could define either `func (f *Foo) method()` or `func (f Foo) method()` if you want reference or value semantics

src/main/scala/temple/generate/server/go/service/GoServiceHookGenerator.scala
```diff
@@ -1,10 +1,56 @@
...
+import scala.collection.immutable.ListMap
+
 object GoServiceHookGenerator {
 
   private[service] def generateImports(root: ServiceRoot): String =
     s"import ${doubleQuote(root.module + "/dao")}"
+
+  private[service] def generateHookStruct(root: ServiceRoot, operations: Set[CRUD]): String = {
+    val beforeCreate = operations.toSeq.sorted.map {
+      case op @ (CRUD.List | CRUD.Read | CRUD.Delete) =>
```
Lewiky, line 19: Was not aware of this syntax until now - pretty fancy

Lewiky (approved):
LGTM

src/main/scala/temple/generate/server/go/service/GoServiceHookGenerator.scala
willjvsmith: neither was I, useful

willjvsmith (approved):
LGTM 🚀

PR #218, opened 2020-04-02
[Go-Map] Update ProjectBuilder to generate Go Auth service

Lewiky:
Adds the calls into the ServerBuilder for building auth services if required by the templefile. 

Also adds all the required go snippets for just the projectbuilder unit tests to pass (they're quite large!) But they are the same as the ones already used by Will in the `GoAuthServiceGenerator` tests, so have in theory already been reviewed. 

Once this is in, I'll either update the E2E tests with a complex example, or try and update *all* of the other builders to handle auth properly


src/main/scala/temple/builder/project/ProjectConfig.scala
```diff
@@ -12,6 +12,7 @@ object ProjectConfig {
   val defaultLanguage: ServiceLanguage = ServiceLanguage.Go
   val defaultDatabase: Database        = Database.Postgres
   val defaultAuth: ServiceAuth         = ServiceAuth.Email
+  val authPort: Int                    = 1023 //* One before the [[Templefile]] start point
```
c-harding: This looks dangerous: 1023 needs sudo, right?

c-harding: This looks dangerous: 1023 needs sudo, right?

c-harding: Also wtf is that comment symbol?

c-harding (approved):
So much go code. Is it identical to that on the spec repo?

src/main/scala/temple/builder/project/ProjectConfig.scala
Lewiky: That is correct... I will update accordingly

src/main/scala/temple/builder/ServerBuilder.scala
```diff
@@ -98,8 +98,8 @@ object ServerBuilder {
       case ServiceAuth.Email =>
         Map(
           "id"       -> Attribute(AttributeType.UUIDType),
+          "email"    -> Attribute(AttributeType.StringType()),
           "password" -> Attribute(AttributeType.StringType()),
-          "email"    -> Attribute(AttributeType.StringType(), valueAnnotations = Set(Annotation.Unique)),
```
jaylees14, line 102: Any reason why this is now not unique?

willjvsmith:
Do we need a separate copy of the snippets if they're identical?

Lewiky:
> Do we need a separate copy of the snippets if they're identical?

@smithwjv Yes, since there is exactly one difference, the package name

src/main/scala/temple/builder/DatabaseBuilder.scala
```diff
@@ -55,6 +55,7 @@ object DatabaseBuilder {
     endpoints: Set[CRUD],
     idAttribute: IDAttribute,
```
jaylees14: Is this variable used now?

jaylees14 (approved):
Lots of Go, but changes look good :)

src/main/scala/temple/builder/ServerBuilder.scala
Lewiky: Yep - this is a read statement so having the restriction that it's unique doesn't actually do anything ;)

src/main/scala/temple/builder/DatabaseBuilder.scala
Lewiky: Nope - removed

PR #219, opened 2020-04-02
[Codegen] Make CodeTermClass a typeclass, to allow for arbitrary nesting

c-harding:


Lewiky (approved):
Black magic but I'll allow it

willjvsmith (approved):
This file just gets worse and worse

jaylees14 (approved):
Looks good :)

PR #220, opened 2020-04-02
[go-gen] Add list handler generation

willjvsmith:
* Adds generation of list handler
* Adds generation of handler stubs
* Adds overloaded functions for `genFunctionCall` and `genMethodCall` to handle optionals


src/main/scala/temple/generate/server/go/service/main/GoServiceMainHandlersGenerator.scala
```diff
@@ -0,0 +1,183 @@
...
+  private def generateListHandler(root: ServiceRoot, responseMap: ListMap[String, String]): String = {
+    // Whether enumerating by created_by field or not
+    val byCreator = root.createdByAttribute match {
+      case CreatedByAttribute.None => false
+      case enumerating: CreatedByAttribute.Enumerating =>
+        enumerating match {
+          case _: CreatedByAttribute.EnumerateByCreator => true
+          case _: CreatedByAttribute.EnumerateByAll     => false
+        }
+    }
```
jaylees14: You can simplify the nested match:
```scala
  val byCreator = root.createdByAttribute match {
      case CreatedByAttribute.None                  => false
      case _: CreatedByAttribute.EnumerateByCreator => true
      case _: CreatedByAttribute.EnumerateByAll     => false
  }
```

src/main/scala/temple/generate/server/go/service/main/GoServiceMainHandlersGenerator.scala
```diff
@@ -0,0 +1,183 @@
...
+
+    mkCode(
```
jaylees14: Can you break this into some variables? It's gonna become really hard to add new things 😢

src/main/scala/temple/generate/server/go/service/main/GoServiceMainHandlersGenerator.scala
```diff
@@ -0,0 +1,183 @@
...
+      ),
+    )
+  }
+
+  /** Generate the create handler function */
+  private def generateCreateHandler(root: ServiceRoot): String =
+    mkCode(
+      generateHandlerDecl(root, Create),
+      CodeWrap.curly.tabbed(
+        ),
```
jaylees14: Interesting formatting, but I'll allow it since you're gonna replace them

src/main/scala/temple/generate/server/go/service/main/GoServiceMainHandlersGenerator.scala
```diff
@@ -0,0 +1,183 @@
...
+      genIfErr(
+        mkCode.lines(
+          generateHTTPError("StatusUnauthorized", "Could not authorize request: %s", genMethodCall("err", "Error")),
+          genReturn(),
+        ),
+      ),
+    )
+
+  /** Generate the list handler function */
+  private def generateListHandler(root: ServiceRoot, responseMap: ListMap[String, String]): String = {
```
jaylees14: If they're all going to get this long, I feel like they should be broken into separate objects?

src/main/scala/temple/generate/server/go/service/main/GoServiceMainHandlersGenerator.scala
```diff
@@ -0,0 +1,183 @@
...
+
+object GoServiceMainHandlersGenerator {
+
+  /** Generate a map for converting the fields of the DAO response to the JSON response  */
+  private def generateResponseMap(root: ServiceRoot): ListMap[String, String] =
+    // Includes ID attribute and all attributes without the @server annotation
+    ListMap(root.idAttribute.name.toUpperCase -> s"${root.name}.${root.idAttribute.name.toUpperCase}") ++
+    root.attributes.collect {
+      case (name, attribute) if !attribute.accessAnnotation.contains(Annotation.Server) =>
+        (
```
jaylees14: Have you thought about using `foo → bar` instead of `(foo, bar)` syntax, I think it reads a lot nicer

src/main/scala/temple/generate/server/go/service/main/GoServiceMainHandlersGenerator.scala
Lewiky:
```suggestion
    val byCreator = root.createdByAttribute match {
        case CreatedByAttribute.EnumerateByCreator(_, _) => true
        case _                                           => false
    }
```

src/main/scala/temple/generate/server/go/service/main/GoServiceMainHandlersGenerator.scala
```diff
@@ -0,0 +1,183 @@
...
+import temple.generate.server.go.common.GoCommonGenerator._
+import temple.utils.StringUtils.doubleQuote
+
+import scala.Option.when
+import scala.collection.immutable.ListMap
+
+object GoServiceMainHandlersGenerator {
+
+  /** Generate a map for converting the fields of the DAO response to the JSON response  */
+  private def generateResponseMap(root: ServiceRoot): ListMap[String, String] =
```
jaylees14, line 28: This function feels like it's a bit too much to be one expression, how about breaking it down?

src/main/scala/temple/generate/server/go/service/main/GoServiceMainHandlersGenerator.scala
```diff
@@ -0,0 +1,183 @@
...
+                genPopulateStruct(s"list${root.name.capitalize}Element", responseMap),
+              ),
+              s"${root.name}ListResp.${root.name.capitalize}List",
+            ),
+          ),
+          genMethodCall(genMethodCall("json", "NewEncoder", "w"), "Encode", s"${root.name}ListResp"),
+        ),
+      ),
+    )
+  }
```
Lewiky: Christ alive - can this be separated out into variables?

src/main/scala/temple/generate/server/go/service/main/GoServiceMainHandlersGenerator.scala
```diff
@@ -0,0 +1,183 @@
...
+      generateHandlerDecl(root, Delete),
+      CodeWrap.curly.tabbed(
+        ),
+    )
+
+  /** Generate the env handler functions */
+  private[service] def generateHandlers(root: ServiceRoot, operations: Set[CRUD]): String = {
+    val responseMap = generateResponseMap(root)
+    mkCode.doubleLines(
+      for (operation <- operations.toSeq.sorted)
```
jaylees14: Any reason why this is a `for` and `yield` instead of a straight forward map?

jaylees14 (commented):
Kinda getting a little hard to read in places, think maybe breaking them down some more would help?

Functionality looks good tho.

src/e2e/resources/go/user/user.go.snippet
```diff
@@ -147,10 +149,44 @@ func jsonMiddleware(next http.Handler) http.Handler {
...
-func (env *env) listTempleuserHandler(w http.ResponseWriter, r *http.Request) {}
+func (env *env) listTempleuserHandler(w http.ResponseWriter, r *http.Request) {
+	templeuserList, err := env.dao.ListTempleuser()
+	if err != nil {
+		errMsg := util.CreateErrorJSON(fmt.Sprintf("Something went wrong: %s", err.Error()))
+		http.Error(w, errMsg, http.StatusInternalServerError)
+		return
+	}
+
+	templeuserListResp := listTempleuserResponse{
```
jaylees14, line 160: Should this not be `listTempleuserResp` if we're following conventions?

jaylees14 (approved):
Nice, much more readable. Thanks for making the changes :)

Lewiky (approved):
Yeah reads nicely 👍

PR #222, opened 2020-04-02
[Go-gen] Update casing in all the generation code

c-harding:
This converts any generated code to use camelCase/PascalCase/snake_case/kebab-case depending on convention for each application, instead of lowercase/Capitalcase

Other effects:

- `FileMatchers`: file diffs are now alphabetical, which should hopefully make it clearer what’s going on when the files don’t match up


src/e2e/scala/temple/SimpleE2ETestData.scala
```diff
@@ -27,19 +27,19 @@ object SimpleE2ETestData {
...
+      |WORKDIR /temple-user
       |
       |COPY go.mod go.sum ./
       |
       |RUN ["go", "mod", "download"]
       |
       |COPY . .
       |
-      |COPY config.json /etc/templeuser-service
+      |COPY config.json /etc/temple-user-service
```
jaylees14: Think your branch might be outdated - this should have a trailing slash from #224

src/main/scala/temple/builder/MetricsBuilder.scala
```diff
@@ -6,6 +6,8 @@ import temple.generate.metrics.grafana.ast.{Datasource, Row}
 
 object MetricsBuilder {
 
+  // TODO: investigate this casing
+  // DON’T LET THIS BE PRed
```
jaylees14: 😱

jaylees14 (approved):
Changes look good - we might find issues later on when we actually come to run it, but nothing really we can do about that now :)

src/test/scala/temple/generate/server/go/GoServiceGeneratorTestData.scala
```diff
@@ -29,8 +29,10 @@ object GoServiceGeneratorTestData {
   )
 
   val simpleServiceFiles: Files = Map(
-    File("user", "go.mod")  -> readFile("src/test/scala/temple/generate/server/go/testfiles/user/go.mod.snippet"),
-    File("user", "user.go") -> readFile("src/test/scala/temple/generate/server/go/testfiles/user/user.go.snippet"),
+    File("user", "go.mod") -> readFile("src/test/scala/temple/generate/server/go/testfiles/user/go.mod.snippet"),
+    File("user", "user.go") -> readFile(
+      "src/test/scala/temple/generate/server/go/testfiles/user/user.go.snippet",
+    ),
```
willjvsmith, line 35: Boosting lines changed 😉

src/main/scala/temple/generate/server/go/common/GoCommonGenerator.scala
```diff
@@ -117,26 +117,17 @@ object GoCommonGenerator {
...
+  private[go] def genFunctionCall(name: String, args: CodeTerm*): String =
     CodeWrap.parens.prefix(name).list(args)
 
   /** Generate function call, but with optional arguments */
   private[go] def genFunctionCall(name: String, firstArg: Option[String], otherArgs: Option[String]*): String =
     CodeWrap.parens.prefix(name).list(firstArg, otherArgs)
 
   /** Generate method call */
-  private[go] def genMethodCall(objectName: String, methodName: String, args: String*): String =
+  private[go] def genMethodCall(objectName: String, methodName: String, args: CodeTerm*): String =
```
willjvsmith, line 128: How come?

willjvsmith (approved):
LGTM 🚀 

It would nice to see more descriptive PR descriptions, bullet pointing everything done and any design decisions made. For example you've improved the map comparison test with `FileMatchers`, something we've been wanting for a while, but it's not mentioned in the desc.

Furthermore, there are other changes being made here that aren't relevant to the PR, but are hidden amongst almost 700 lines of changes.

src/main/scala/temple/generate/server/go/common/GoCommonGenerator.scala
c-harding: This was because my last PR #219 had enabled the nesting that was needed here so this could be simplified, and I spotted it when I was checking this directory for renaming

src/test/scala/temple/generate/server/go/GoServiceGeneratorTestData.scala
c-harding: That was the autoformatter!

PR #223, opened 2020-04-02
[go-gen][metrics] Extract shared logic to common generator

jaylees14:
Quite a lot of logic will be shared between `Service` and `Auth` metrics, so I've moved that stuff into the `common` folder.

No other logic changes


Lewiky (approved):
Sick

willjvsmith (approved):
LGTM 🚀

PR #224, opened 2020-04-02
[docker] Add trailing slash to path

jaylees14:
Turns out the trailing slash is quite important... otherwise you don't copy to the directory 🤦‍♂


c-harding (approved):
///

Lewiky (approved):

PR #225, opened 2020-04-03
Auth metric gen

jaylees14:
Pretty easy one - generate metrics for `auth` service. They differ from the main metrics in the globals exposed: for standard services we have:

```
var RequestCreate = "create"
...
```
For auth services, these are replaced with:

```
var RequestRegister = "register"
var RequestLogin      = "login"
```

The tests files were taken directly from `spec-golang`, so pretty confident the generation is correct :)


Lewiky (approved):
Pretty easy - looks good. Does the `auth.go` file not also need to change to support metrics?

jaylees14:
> Pretty easy - looks good. Does the `auth.go` file not also need to change to support metrics?

Yes indeed, that's next :)

c-harding (approved):

PR #226, opened 2020-04-03
[Go-Map] Update E2E tests to check for generated auth go files

Lewiky:
As title says really - The go files here are exactly the same as the ProjectBuilder unit tests, but because they're in different packages (e2e vs test) they have different resources folders so need to be copied across, but can turn off the .snippet file diffs


jaylees14 (approved):
LGTM, nice one

c-harding (approved):
if this matches the spec repo, good to go

jaylees14 (approved):
LGTM <3

PR #227, opened 2020-04-03
[go-gen][hook] Move common hook logic into shared package

jaylees14:
Same as #223 but this time for hooks. Move all the common logic into the `common` package, and call this from the `GoAuthServiceGenerator`. Will add the auth specific stuff in the next PR.

Updates tests etc.


willjvsmith (approved):
Standard, LGTM 🚀

c-harding (approved):

Lewiky (approved):
Lovin it

PR #228, opened 2020-04-03
[go-gen] Fix foreign keys being passed in as UUID type

willjvsmith:
* Changes foreign keys to be passed in as the foreign key type, rather than the UUID type
  * Needed to switch on, so as to know which inter-service communication function to call when checking incoming foreign key parameters in the create handlers

This should remove the assumption that UUIDs in the user-defined attributes are always foreign keys, but there might be a check left in that I can't find. I need this change to carry on handler generation :)

### DIFF
--- a/src/main/scala/temple/builder/ServerBuilder.scala
+++ b/src/main/scala/temple/builder/ServerBuilder.scala
@@ -66,12 +66,6 @@ object ServerBuilder {
       case (_, Attribute(x: ForeignKey, _, _)) => x.references
     }.toSeq
 
-    // In the server code all foreign keys are stored as UUID types - map into the correct type
-    val attributes: ListMap[String, Attribute] = ListMap.from(serviceBlock.attributes.map {
-      case (str, Attribute(_: ForeignKey, access, value)) => (str, Attribute(AttributeType.UUIDType, access, value))
-      case default                                        => default
-    })
-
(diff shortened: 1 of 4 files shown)


jaylees14 (approved):
Looks good - obviously can't check everywhere, but if the compiler's happy you're probably good

src/main/scala/temple/builder/ServerBuilder.scala
```diff
@@ -66,12 +66,6 @@ object ServerBuilder {
       case (_, Attribute(x: ForeignKey, _, _)) => x.references
     }.toSeq
 
-    // In the server code all foreign keys are stored as UUID types - map into the correct type
-    val attributes: ListMap[String, Attribute] = ListMap.from(serviceBlock.attributes.map {
-      case (str, Attribute(_: ForeignKey, access, value)) => (str, Attribute(AttributeType.UUIDType, access, value))
-      case default                                        => default
-    })
-
     // TODO: Auth
```
willjvsmith, line 69: Should the TODO still be here @Lewiky ?

Lewiky (approved):

src/main/scala/temple/builder/ServerBuilder.scala
Lewiky: Well it's being done now!!

PR #229, opened 2020-04-03
[config] Add config JSON generator

jaylees14:
Add a generator for `config.json` from spec-golang.

Nothing too exciting - will do the mapping after because of all tests that will need updating

### DIFF
--- a/src/main/scala/temple/generate/server/config/ServerConfigGenerator.scala
+++ b/src/main/scala/temple/generate/server/config/ServerConfigGenerator.scala
@@ -1,0 +1,16 @@
+package temple.generate.server.config
+
+import io.circe.syntax._
+import temple.ast.Metadata.Database
+import temple.ast.Templefile.Ports
+import temple.generate.server.config.ast.{PostgresConfig, ServerConfig}
+
+object ServerConfigGenerator {
+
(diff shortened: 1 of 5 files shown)


src/main/scala/temple/generate/server/config/ServerConfigGenerator.scala
```diff
@@ -0,0 +1,16 @@
...
+import io.circe.syntax._
+import temple.ast.Metadata.Database
+import temple.ast.Templefile.Ports
+import temple.generate.server.config.ast.{PostgresConfig, ServerConfig}
+
+object ServerConfigGenerator {
+
+  def generate(serviceName: String, database: Database, services: Map[String, String], ports: Ports): String = {
+    val databaseConfig = database match {
+      case Database.Postgres => PostgresConfig(serviceName + "-db")
```
jaylees14, line 12: @Lewiky this is what you do in Kube right?

c-harding (approved):
nice and simple

src/main/scala/temple/generate/server/config/ServerConfigGenerator.scala
Lewiky: Yeah bruh

Lewiky (approved):
Nice

PR #230, opened 2020-04-03
[Auth-Map] Update projectBuilder targets to generate auth services 

Lewiky:
Updates the whole ProjectBuilder to make sure auth service is generated by all sub-builders when appropriate

also updates the tests, which means doubling all the generated output (as the complex user test uses auth)


src/main/scala/temple/builder/OrchestrationBuilder.scala
```diff
@@ -36,7 +36,7 @@ object OrchestrationBuilder {
             dbLifecycleCommand = dbLanguage match {
               case Database.Postgres => LifecycleCommand.psqlSetup.toString
             },
-            usesAuth = true,
+            usesAuth = if (name == "auth") false else true,
```
jaylees14: Gives me confidence for the next 1124 lines
```suggestion
            usesAuth = name != "auth"
```

src/main/scala/temple/builder/project/ProjectBuilder.scala
```diff
@@ -53,30 +53,56 @@ object ProjectBuilder {
...
+              val postgresStatements                = createStatements.map(PostgresGenerator.generate).mkString("\n\n")
+              (File(s"${name.toLowerCase}-db", "init.sql"), postgresStatements)
+          }
+      } ++ Option.when(usesAuth) {
+        templefile.lookupMetadata[Database].getOrElse(ProjectConfig.defaultDatabase) match {
           case Database.Postgres =>
-            implicit val context: PostgresContext = PostgresContext(QuestionMarks)
-            val postgresStatements                = createStatements.map(PostgresGenerator.generate).mkString("\n\n")
-            (File(s"${name.toLowerCase}-db", "init.sql"), postgresStatements)
+            implicit val postgresContext: PostgresContext = PostgresContext(PreparedType.DollarNumbers)
```
jaylees14: Why `DollarNumbers` here and `QuestionMarks` above?

src/main/scala/temple/builder/OrchestrationBuilder.scala
c-harding: wat

src/main/scala/temple/builder/project/ProjectBuilder.scala
```diff
@@ -53,30 +53,56 @@ object ProjectBuilder {
...
+    val authServiceBlock: ServiceBlock = ServiceBlock(
+      attributes = Map(
+        "id"       -> Attribute(AttributeType.UUIDType, valueAnnotations = Set(Annotation.Unique)),
+        "email"    -> Attribute(AttributeType.StringType(), valueAnnotations = Set(Annotation.Unique)),
+        "password" -> Attribute(AttributeType.StringType()),
+      ),
+    )
+    authServiceBlock.setParent(templefile)
+
     val databaseCreationScripts = templefile.services.map {
```
jaylees14: Surely something like this would be nicer to read, rather than duplicating all this code

```
val services = templefile.services + (if (usesAuth) ("auth", authServiceBlock) else ())
```

src/main/scala/temple/builder/project/ProjectBuilder.scala
```diff
@@ -53,30 +53,56 @@ object ProjectBuilder {
...
+          val createStatements: Seq[Statement.Create] = DatabaseBuilder.createServiceTables(name, service)
+          service.lookupMetadata[Database].getOrElse(ProjectConfig.defaultDatabase) match {
+            case Database.Postgres =>
+              implicit val context: PostgresContext = PostgresContext(QuestionMarks)
+              val postgresStatements                = createStatements.map(PostgresGenerator.generate).mkString("\n\n")
+              (File(s"${name.toLowerCase}-db", "init.sql"), postgresStatements)
+          }
+      } ++ Option.when(usesAuth) {
+        templefile.lookupMetadata[Database].getOrElse(ProjectConfig.defaultDatabase) match {
           case Database.Postgres =>
```
c-harding: I love how these two lines say a grand total of nothing right now

src/main/scala/temple/builder/project/ProjectBuilder.scala
jaylees14: Or if `templefile` could expose a way to add additional services?

jaylees14 (commented):
Couple of things, quite a lot of repeated code. Wonder if we could expose a new method on `Templefile` that adds a new service?

c-harding (approved):
big but looks good
edit: e2e is failing

src/main/scala/temple/ast/Templefile.scala
```diff
@@ -19,7 +20,24 @@ case class Templefile(
   def servicesWithPorts: Iterable[(String, ServiceBlock, Ports)] =
     services
       .zip(Iterator.from(1025, step = 2)) // 1024 is reserved for auth service
-      .map { case ((name, service), port) => (name, service, Ports(port, port + 1)) }
+      .map {
+        case ((name, service), port) =>
+          if (name == "Auth") (name, service, Ports(ProjectConfig.authPort, ProjectConfig.authPort + 1))
```
jaylees14: Hmm I'm not really a fan of this string comparison, especially given that there's no `.lowercase` or such. How about creating an `AuthServiceBlock` subclass and then you match on the types?

src/main/scala/temple/ast/Templefile.scala
```diff
@@ -19,7 +20,24 @@ case class Templefile(
   def servicesWithPorts: Iterable[(String, ServiceBlock, Ports)] =
     services
       .zip(Iterator.from(1025, step = 2)) // 1024 is reserved for auth service
```
jaylees14: Hmm - surely 1025 is `ProjectConfig.authPort + 1`, which is used below? I think all of these should be coming from the same place - bring this into a `ProjectConfig` variable too?

src/main/scala/temple/ast/Templefile.scala
jaylees14: Also, how about bringing `ProjectConfig.authPort + 1` into its own variable in `ProjectConfig`?

src/main/scala/temple/ast/Templefile.scala
```diff
@@ -19,7 +20,24 @@ case class Templefile(
...
-      .map { case ((name, service), port) => (name, service, Ports(port, port + 1)) }
+      .map {
+        case ((name, service), port) =>
+          if (name == "Auth") (name, service, Ports(ProjectConfig.authPort, ProjectConfig.authPort + 1))
+          else (name, service, Ports(port, port + 1))
+      }
+
+  def servicesWithPortsWithoutAuth: Iterable[(String, ServiceBlock, Ports)] =
```
jaylees14: How about `providedServicesWithPorts` and `allServicesWithPorts` - I think without sounds a bit verbose?

jaylees14 (commented):
Hmm, I'm not sure about comparing strings - I probably could have given a better description when I suggested it, sorry.

src/main/scala/temple/ast/Templefile.scala
```diff
@@ -9,17 +10,42 @@ case class Templefile(
...
+    */
+  def providedServices: Map[String, ServiceBlock] = services.collect {
+    case (name: String, service: ServiceBlock) => name -> service
+  }
+
+  def allServicesWithPorts: Iterable[(String, GeneratedBlock, Ports)] =
     services
-      .zip(Iterator.from(1025, step = 2)) // 1024 is reserved for auth service
-      .map { case ((name, service), port) => (name, service, Ports(port, port + 1)) }
+      .zip(Iterator.from(ProjectConfig.serviceStartPort, step = 2)) // 1024 is reserved for auth service
```
jaylees14:
```suggestion
      .zip(Iterator.from(ProjectConfig.serviceStartPort, step = 2))
```

src/main/scala/temple/builder/OpenAPIBuilder.scala
```diff
@@ -11,17 +11,20 @@ object OpenAPIBuilder {
...
-            operations = endpoints(block),
-            attributes = block.attributes,
-            structs = block.structs.map {
-              case (structName, structBlock) =>
-                Service.Struct(structName, endpoints(structBlock), attributes = structBlock.attributes)
-            },
-          )
-      }.toSeq,
+      services = templefile.services
+        .filterNot(_._1 == "Auth")
```
jaylees14: Oooh can we use one of the new `.service` methods?

jaylees14 (approved):
Nice one, just 2 small things I spotted but otherwise good to go :)

src/main/scala/temple/ast/AuthServiceBlock.scala
```diff
@@ -0,0 +1,15 @@
...
+case object AuthServiceBlock extends GeneratedBlock {
+
+  override def attributes: Map[String, Attribute] = Map(
+    "id"       -> Attribute(AttributeType.UUIDType, valueAnnotations = Set(Annotation.Unique)),
+    "email"    -> Attribute(AttributeType.StringType(), valueAnnotations = Set(Annotation.Unique)),
+    "password" -> Attribute(AttributeType.StringType()),
+  )
+
+  override def metadata                          = Seq(Omit(Set(Endpoint.Update, Endpoint.Delete)))
+  override def structs: Map[String, StructBlock] = Map()
```
c-harding: Can these become `val`s? I’m not an expert at how the JVM is optimised, but I would imagine it beneficial if these were statically allocated

src/main/scala/temple/ast/Templefile.scala
```diff
@@ -9,17 +10,42 @@ case class Templefile(
...
 ) extends TempleNode {
   // Inform every child node of their parent, so that they can access the project information
   for (block <- Iterator(projectBlock) ++ targets.valuesIterator ++ services.valuesIterator) {
     block.setParent(this)
   }
 
-  def servicesWithPorts: Iterable[(String, ServiceBlock, Ports)] =
+  /**
+    * The services provided in the templefile, i.e the ServiceBlocks, not the AuthServiceBlocks
+    * @return
```
c-harding: Thanks, glad to know it returns

src/main/scala/temple/ast/GeneratedBlock.scala
```diff
@@ -0,0 +1,25 @@
+package temple.ast
+
+import temple.ast.Metadata.ServiceMetadata
+
+trait GeneratedBlock extends TempleBlock[ServiceMetadata] {
```
c-harding: I’m tempted to say that this should be in the same file as its two children, so it can be sealed

src/main/scala/temple/ast/Templefile.scala
```diff
@@ -9,17 +10,42 @@ case class Templefile(
...
+
+  def allServicesWithPorts: Iterable[(String, GeneratedBlock, Ports)] =
     services
-      .zip(Iterator.from(1025, step = 2)) // 1024 is reserved for auth service
-      .map { case ((name, service), port) => (name, service, Ports(port, port + 1)) }
+      .zip(Iterator.from(ProjectConfig.serviceStartPort, step = 2)) // 1024 is reserved for auth service
+      .map {
+        case ((name, service), port) =>
+          if (service == AuthServiceBlock) (name, service, Ports(ProjectConfig.authPort, ProjectConfig.authMetricPort))
+          else (name, service, Ports(port, port + 1))
```
c-harding: This feels a little silly that the auth service is assigned a port, but we don’t use it

src/main/scala/temple/ast/GeneratedBlock.scala
c-harding: And `GeneratedBlock` sounds to me like it only encompasses the blocks that aren’t provided, like the auth block

src/main/scala/temple/builder/OpenAPIBuilder.scala
c-harding: `providedServices` seems like what we want here, and it’s less stringly typed

src/main/scala/temple/builder/OrchestrationBuilder.scala
```diff
@@ -36,7 +36,7 @@ object OrchestrationBuilder {
             dbLifecycleCommand = dbLanguage match {
               case Database.Postgres => LifecycleCommand.psqlSetup.toString
             },
-            usesAuth = true,
+            usesAuth = name != "Auth",
```
c-harding: `service != AuthServiceBlock`? And how does this work if the project doesn’t have an auth block?

src/main/scala/temple/ast/Templefile.scala
```diff
@@ -9,17 +10,42 @@ case class Templefile(
...
+    allServicesWithPorts.collect { case (name, service: ServiceBlock, ports) => (name, service, ports) }
+
+  /**
+    * Add an extra service to an exisiting Templefile - useful for adding auth when needed.
+    * @param name - the name of the new service
+    * @param block - the [[ServiceBlock]] to add
+    */
+  def addService(name: String, block: GeneratedBlock): Unit = {
+    block.setParent(this)
+    services += name -> block
```
c-harding: Please no, case classes really shouldn’t have modifiable parameters

c-harding: and if we’re setting the parent of AuthServiceBlock, then we’ll need to change AuthServiceBlock to a (case) class rather than a case object, otherwise that parent will persist (including across test runs etc).

c-harding: One solution (and stops a port being "wasted" on auth):

Keep a constructor argument `providedServices: Map[String, ServiceBlock] = Map()`, then use these as attributes to Templefile

```scala
// Whether or not to generate an auth service - based on whether any service has #auth
private val usesAuth = templefile.services.exists {
  case (_, service) => service.lookupMetadata[ServiceAuth].nonEmpty
}

val providedServicesWithPorts: Iterable[(String, ServiceBlock, Ports)] =
  providedServices
    .zip(Iterator.from(ProjectConfig.serviceStartPort, step = 2))
    .map { case ((name, service), ports) => (name, service, ports) }

val allServicesWithPorts: Iterable[(String, GeneratedBlock, Int)] = {
  if (usesAuth) {
    val authBlock = AuthServiceBlock()
    authBlock.setParent(this)
    Iterable((auth, authBlock, Ports(ProjectConfig.authPort, ProjectConfig.authMetricPort))) ++ providedServicesWithPorts
  } else providedServicesWithPorts

val allServices: Map[String, GeneratedBlock] = allServicesWithPorts.map { case (name, service, ports) => name, service) }.toMap
```

src/main/scala/temple/builder/OrchestrationBuilder.scala
c-harding: Is there a reason why `createServiceOrchestrationRoot` doesn’t take a Templefile? That would allow us to just call `Templefile.usesAuth`

src/main/scala/temple/builder/project/ProjectBuilder.scala
```diff
@@ -49,22 +49,28 @@ object ProjectBuilder {
...
   def build(templefile: Templefile, detail: LanguageDetail): Project = {
 
+    // Whether or not to generate an auth service - based on whether any service has #auth
     val usesAuth = templefile.services.exists {
       case (_, service) => service.lookupMetadata[ServiceAuth].nonEmpty
     }
 
+    if (usesAuth) {
+      templefile.addService("Auth", AuthServiceBlock)
+    }
```
c-harding: This seems very dangerous, see comments above

c-harding (changes requested):
Some concerns with the use of mutable parameters to Templefile. I’ve included a suggested alternative solution

src/main/scala/temple/builder/OrchestrationBuilder.scala
Lewiky: this usesAuth is a service level thing, it's whether this particular service needs auth in order to be accessed, so using the whole templefile doesn't do what we need

c-harding (approved):
lgtm

PR #221, opened 2020-04-02
[go-gen][hooks] Introduce AddHook methods

jaylees14:
Generate the utility methods for adding hooks to the list of hooks.

Update allllll the tests

PR #231, opened 2020-04-03
[go-gen][auth][hook] Generate hooks for auth service

jaylees14:
Let's have some hooks in the `auth` service! Similar to #221, but the `Auth` service has some custom names / types, so has to be separate


src/main/scala/temple/generate/server/go/auth/GoAuthServiceHookGenerator.scala
```diff
@@ -0,0 +1,68 @@
...
+        s"before${action}Hooks" -> s"[]*${generateBeforeHookType(action, op)}"
+    }
+
+    val afterCreate = endpoints.map {
+      case (action, _) =>
+        s"after${action}Hooks" -> s"[]*${generateAfterHookType}"
+    }
+
+    mkCode.lines(
+      "// Hook allows additional code to be executed before and after every datastore interaction",
```
willjvsmith: could be dragged out to common but 🤷‍♂

willjvsmith (approved):
LGTM 🚀

Lewiky (approved):

PR #232, opened 2020-04-03
[go-gen] Add generation of create handler

willjvsmith:
* Adds generation of create handlers
  * Splits up generating functions ready for reuse with update handler
* Adds GoHTTPStatus enum
* Updates tests

Should be pretty readable, will make it reasonably quick to write the update handler. About 300 lines of snippets.

A few issues that will need to be handled:
  * What if there are no client facing attributes
  * ServerSet values cannot be created in the SQL query

I have made cards for these issues.

### DIFF
--- a/src/e2e/resources/go/user/user.go.snippet
+++ b/src/e2e/resources/go/user/user.go.snippet
@@ -180,7 +180,72 @@ func (env *env) listTempleuserHandler(w http.ResponseWriter, r *http.Request) {
 }
 
 func (env *env) createTempleuserHandler(w http.ResponseWriter, r *http.Request) {
+	auth, err := util.ExtractAuthIDFromRequest(r.Header)
+	if err != nil {
+		errMsg := util.CreateErrorJSON(fmt.Sprintf("Could not authorize request: %s", err.Error()))
+		http.Error(w, errMsg, http.StatusUnauthorized)
+		return
+	}
 
+	var req createTempleuserRequest
+	err = json.NewDecoder(r.Body).Decode(&req)
+	if err != nil {
+		errMsg := util.CreateErrorJSON(fmt.Sprintf("Invalid request parameters: %s", err.Error()))
+		http.Error(w, errMsg, http.StatusBadRequest)
+		return
+	}
+
+	if req.Username == nil || req.Email == nil || req.FirstName == nil || req.LastName == nil || req.CreatedAt == nil || req.NumberOfDogs == nil || req.CurrentBankBalance == nil || req.BirthDate == nil || req.BreakfastTime == nil {
+		errMsg := util.CreateErrorJSON("Missing request parameter(s)")
+		http.Error(w, errMsg, http.StatusBadRequest)
+		return
+	}
+
+	_, err = valid.ValidateStruct(req)
+	if err != nil {
+		errMsg := util.CreateErrorJSON(fmt.Sprintf("Invalid request parameters: %s", err.Error()))
+		http.Error(w, errMsg, http.StatusBadRequest)
+		return
+	}
+
+	uuid, err := uuid.NewUUID()
+	if err != nil {
+		errMsg := util.CreateErrorJSON(fmt.Sprintf("Could not create UUID: %s", err.Error()))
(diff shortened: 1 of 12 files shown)


src/e2e/resources/go/user/user.go.snippet
```diff
@@ -180,7 +180,72 @@ func (env *env) listTempleuserHandler(w http.ResponseWriter, r *http.Request) {
...
+
+	var req createTempleuserRequest
+	err = json.NewDecoder(r.Body).Decode(&req)
+	if err != nil {
+		errMsg := util.CreateErrorJSON(fmt.Sprintf("Invalid request parameters: %s", err.Error()))
+		http.Error(w, errMsg, http.StatusBadRequest)
+		return
+	}
+
+	if req.Username == nil || req.Email == nil || req.FirstName == nil || req.LastName == nil || req.CreatedAt == nil || req.NumberOfDogs == nil || req.CurrentBankBalance == nil || req.BirthDate == nil || req.BreakfastTime == nil {
```
jaylees14, line 198: Fantastic 😆

src/main/scala/temple/generate/server/go/service/main/GoServiceMainCreateHandlerGenerator.scala
```diff
@@ -1,16 +1,89 @@
...
+        mkCode.doubleLines(
+          generateExtractAuthBlock(),
+          generateDecodeRequestBlock(s"create${root.name.capitalize}"),
+          generateRequestNilCheck(root, clientAttributes),
+          generateValidateStructBlock(),
+          when(usesComms) { generateForeignKeyCheckBlocks(root, clientAttributes) },
+          when(!hasAuthBlock) { generateNewUUIDBlock() },
+          generateDAOCallBlock(root, hasAuthBlock, clientAttributes),
+          generateJSONResponse(s"create${root.name.capitalize}", responseMap),
+        ),
```
jaylees14, line 86: Nice, very readable

src/main/scala/temple/generate/server/go/service/main/GoServiceMainHandlersGenerator.scala
```diff
@@ -56,25 +57,110 @@ object GoServiceMainHandlersGenerator {
...
+      ),
+    )
+
+  /** Generate the block for checking foreign keys against other services */
+  private[main] def generateForeignKeyCheckBlocks(
+    root: ServiceRoot,
+    clientAttributes: ListMap[String, Attribute],
+  ): String =
+    mkCode.doubleLines(
+      clientAttributes.map {
```
jaylees14, line 134: Feels like it could be a `collect`?

jaylees14 (approved):
LGTM, nice stuff

src/e2e/resources/go/user/user.go.snippet
willjvsmith: #readable 😉

src/main/scala/temple/generate/server/go/service/main/GoServiceMainHandlersGenerator.scala
willjvsmith: would that be any better considering the `reference` needs to be passed to the function?

jaylees14: I just meant it from the perspective of removing the last case: `case _ => ""`

jaylees14: Something like 

```scala
clientAttributes.view.mapValues(_.attributeType).collect {
    case name -> AttributeType.ForeignKey(reference) => generateForeignKeyCheckBlock(root, name, reference)
},
```

c-harding: unfortunately it’s double-nested, so can’t simply turn the outer one to `collect`. Would instead have to fully destructure attribute `case name -> Attribute(AttributeType.ForeignKey(reference),_,_)`, not sure if that’s nicer

c-harding (approved):
nice

PR #233, opened 2020-04-03
[go-gen] Add readable/writable to ServiceRoot

willjvsmith:
* Adds readable and writable fields to `ServiceRoot`

We decided the defaults are as follows:
  * If any service in the project has an auth block, readable and writable default to `(by: this)`
  * If no service in the project has an auth block, readable and writable default to `(by: all)`

This does bring up that if no service has an auth block, we cannot read or write `(by: this)`, and so therefore need to be blocked at the DSL level. I have made cards, for Charlie, for this.


willjvsmith:
Thoughts on baking readable and writable into type with CreatedBy?

c-harding (approved):
looks good

jaylees14 (approved):
LGTM, nice and clean

PR #234, opened 2020-04-03
[Auth-Map] Add GeneratedBlock trait for more abstraction

Lewiky:
Adds the GeneratedBlock trait, which abstracts above a ServiceBlock, to allow pattern matching between ServiceBlock and AuthServiceBlock


jaylees14 (approved):
Beauty <3

PR #235, opened 2020-04-04
[Validator] Perform renaming of structs/services and attributes (just the tests)

c-harding:
Add tests to #236 to fully demonstrate the effects of resolving naming clashes.

## Original PR message

Yet another painfully large PR, mainly because of the size of the tests. In checking that the renaming was correctly performed, I slightly underestimated the effects of making a change to simple.temple on the e2e test files.

The test changes are a result of renaming a service to `User`, an attribute to `user`, and another service to `Group`, all of which would lead to parse errors in Postgres. As such, they get renamed by adding the project name (SimpleTempleTest) as a prefix.

### Diff breakdown
- **+10–10** in `test` (this is just renaming in the Templefile)
- **+377–369** in `e2e` (this is just renaming the stuff generated from the Templefile)
- **+331–32** in `main`
  - **+138** in `NameClashes`, a new object to handle names reserved by Postgres. This is composable so any lists of names can be used to determine conflicts. **100** lines of this are just a list of Postgres reserved keywords
  - **–2** in `SemanticContext`, removing that this was a function (see `Validator`).
  - **+88** in `ServiceRenamer`, a new class for performing renaming (of services/structs) on a Templefile. It is constructed with a map from old names to new, and then serves as a function from Templefile to Templefile. A lot of this file is exhaustive matches, even when they are just the identity function. This is so that any newly added metadata types etc. will throw MatchErrors here instead of silently allowing unrenamed items through, which would become a real pain to debug.
  - **+102–30** in `Validator`. The validator now also performs renaming on the Templefile’s fields and attributes (Suggestions of what to rename this class (and its methods) to are welcome!).
    It seemed a bit strange to me that `context` was a function, so now an explicit `foreachValueWithContext` etc is used, serving basically the same purpose but a little more explicitly.


jaylees14 (dismissed):
Damn, quite a lot of changes if you hit a reserved keyword. Nice.

jaylees14 (approved):
Much change

Lewiky (approved):
I really think we should strongly encourage users against calling a service "User"

PR #236, opened 2020-04-04
[Validator] Perform renaming of structs/services and attributes (without test update)

c-harding:
Splitting up #235 by removing all the updates to tests. Everything still passes, it just doesn’t demonstrate any of the pwer of this PR. Check out #235 to see the full gory detail of what happens when you apply it to tests, but basically if any naming conflict is found, it is then resolved by prefixing the name with the project name (repeatedly until no more conflicts).

- `NameClashes`, a new object to handle names reserved by Postgres. This is composable so any lists of names can be used to determine conflicts. **100** lines of this are just a list of Postgres reserved keywords
- `SemanticContext`, removing that this was a function (see `Validator`).
- `ServiceRenamer`, a new class for performing renaming (of services/structs) on a Templefile. It is constructed with a map from old names to new, and then serves as a function from Templefile to Templefile. A lot of this file is exhaustive matches, even when they are just the identity function. This is so that any newly added metadata types etc. will throw MatchErrors here instead of silently allowing unrenamed items through, which would become a real pain to debug.
- `Validator`: The validator now also performs renaming on the Templefile’s fields and attributes (Suggestions of what to rename this class (and its methods) to are welcome!).
  It seemed a bit strange to me that `context` was a function, so now an explicit `foreachValueWithContext` etc is used, serving basically the same purpose but a little more explicitly.

Re codecov, every new untested line is tested in the e2e tests

Spurious branch naming as I cherry-picked my way to two non-XXL PRs.

### DIFF
--- a/src/main/scala/temple/DSL/semantics/NameClashes.scala
+++ b/src/main/scala/temple/DSL/semantics/NameClashes.scala
@@ -1,0 +1,148 @@
+package temple.DSL.semantics
+
+import temple.utils.StringUtils
+
+import scala.collection.Iterator.iterate
+
+object NameClashes {
+
+  // https://www.postgresql.org/docs/10/sql-keywords-appendix.html
+  // only those marked as "reserved"/"reserved (can be function or type)" in Postgres
+  private val postgresNames = Set(
+    "all",
+    "analyse",
+    "analyze",
+    "and",
+    "any",
+    "array",
+    "as",
+    "asc",
+    "asymmetric",
+    "authorization",
+    "binary",
+    "both",
+    "case",
+    "cast",
+    "check",
+    "collate",
+    "collation",
+    "column",
+    "concurrently",
+    "constraint",
+    "create",
+    "cross",
+    "current_catalog",
+    "current_date",
+    "current_role",
+    "current_schema",
+    "current_time",
+    "current_timestamp",
+    "current_user",
+    "default",
+    "deferrable",
+    "desc",
+    "distinct",
+    "do",
+    "else",
+    "end",
+    "except",
+    "false",
+    "fetch",
+    "for",
+    "foreign",
+    "freeze",
+    "from",
+    "full",
+    "grant",
+    "group",
+    "having",
+    "ilike", // sql, it’s my favourite language
+    "in",
+    "initially",
+    "inner",
+    "intersect",
+    "into",
+    "is",
+    "isnull",
+    "join",
+    "lateral",
+    "leading",
+    "left",
+    "like",
+    "limit",
+    "localtime",
+    "localtimestamp",
+    "natural",
+    "not",
+    "notnull",
+    "null",
+    "offset",
+    "on",
+    "only",
+    "or",
+    "order",
+    "outer",
+    "overlaps",
+    "placing",
+    "primary",
+    "references",
+    "returning",
+    "right",
+    "select",
+    "session_user",
+    "similar",
+    "some",
+    "symmetric",
+    "table",
+    "tablesample",
+    "then",
+    "to",
+    "trailing",
+    "true",
+    "union",
+    "unique",
+    "user",
+    "using",
+    "variadic",
+    "verbose",
+    "when",
+    "where",
+    "window",
+    "with",
+  )
+
+  case class NameValidator private (private[NameClashes] val isValid: (String, Set[String]) => Boolean) {
+
+    def &(other: NameValidator): NameValidator = NameValidator { (str, taken) =>
+      this.isValid(str, taken) && other.isValid(str, taken)
+    }
+
+    def join(others: IterableOnce[NameValidator]): NameValidator = others.iterator.fold(this) { _ & _ }
+  }
+
+  private def iterableValidator(strings: Set[String]): NameValidator = NameValidator { (name, taken) =>
+    !strings.iterator.contains(name) && !taken.contains(name)
+  }
+
+  private def caseInsensitiveIterableValidator(strings: Set[String]): NameValidator = NameValidator { (name, taken) =>
+    val lowerName = name.toLowerCase
+    !strings.iterator.contains(lowerName) && !taken.contains(lowerName)
+  }
+
+  val postgresValidator: NameValidator = caseInsensitiveIterableValidator(postgresNames)
+
+  def nameSuggestions(name: String, project: String, decapitalize: Boolean = false): Iterator[String] =
+    iterate(name.capitalize) { project + _ }.map {
+      if (decapitalize) StringUtils.decapitalize
+      else identity
+    }
+
+  def dodgeNames(name: String, project: String, takenNames: Set[String], decapitalize: Boolean = false)(
+    nameValidator1: NameValidator,
+    nameValidators: NameValidator*,
+  ): String = {
+    val nameValidator = nameValidator1.join(nameValidators)
+    nameSuggestions(name, project, decapitalize).filter { nameValidator.isValid(_, takenNames) }.next
+  }
+
+}
--- a/src/main/scala/temple/DSL/semantics/NameClashes.scala
+++ b/src/main/scala/temple/DSL/semantics/NameClashes.scala
@@ -111,38 +111,31 @@ object NameClashes {
     "with",
   )
 
-  case class NameValidator private (private[NameClashes] val isValid: (String, Set[String]) => Boolean) {
+  case class NameValidator private (private[NameClashes] val isValid: String => Boolean)
 
-    def &(other: NameValidator): NameValidator = NameValidator { (str, taken) =>
-      this.isValid(str, taken) && other.isValid(str, taken)
+  private[NameClashes] object NameValidator {
+    private val acceptEverything = NameValidator(_ => true)
+
+    def combine(validators: IterableOnce[NameValidator]): NameValidator =
+      validators.iterator.foldLeft(acceptEverything) { (v1, v2) =>
+        NameValidator(str => v1.isValid(str) && v2.isValid(str))
+      }
+
+    def fromIterable(strings: Set[String]): NameValidator = NameValidator { name =>
+      !strings.iterator.contains(name.toLowerCase)
     }
-
-    def join(others: IterableOnce[NameValidator]): NameValidator = others.iterator.fold(this) { _ & _ }
   }
 
-  private def iterableValidator(strings: Set[String]): NameValidator = NameValidator { (name, taken) =>
-    !strings.iterator.contains(name) && !taken.contains(name)
-  }
+  val postgresValidator: NameValidator = NameValidator.fromIterable(postgresNames)
 
-  private def caseInsensitiveIterableValidator(strings: Set[String]): NameValidator = NameValidator { (name, taken) =>
-    val lowerName = name.toLowerCase
-    !strings.iterator.contains(lowerName) && !taken.contains(lowerName)
-  }
-
-  val postgresValidator: NameValidator = caseInsensitiveIterableValidator(postgresNames)
-
+  /** Repeatedly add the project name to the start of the string */
   def nameSuggestions(name: String, project: String, decapitalize: Boolean = false): Iterator[String] =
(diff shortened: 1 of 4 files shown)


src/main/scala/temple/DSL/semantics/NameClashes.scala
```diff
@@ -0,0 +1,141 @@
...
+
+    def fromIterable(strings: Set[String]): NameValidator = NameValidator { name =>
+      !strings.iterator.contains(name.toLowerCase)
+    }
+  }
+
+  val postgresValidator: NameValidator = NameValidator.fromIterable(postgresNames)
+
+  /** Repeatedly add the project name to the start of the string */
+  def nameSuggestions(name: String, project: String, decapitalize: Boolean = false): Iterator[String] =
```
jaylees14: This can be private, no?

src/main/scala/temple/DSL/semantics/NameClashes.scala
```diff
@@ -0,0 +1,141 @@
...
+    iterate(name.capitalize) { project + _ }.map { if (decapitalize) StringUtils.decapitalize else identity }
+
+  def dodgeNames(name: String, project: String, takenNames: Set[String], decapitalize: Boolean = false)(
+    validators: NameValidator*,
+  ): String = {
+    val validator = NameValidator.combine(Iterator(NameValidator.fromIterable(takenNames)) ++ validators)
+    nameSuggestions(name, project, decapitalize).filter { validator.isValid(_) }.next
```
jaylees14: I had to look up how this was implemented to double check it was efficient - from first glance you're filtering over an infinite iterator, which didn't seem right 😆

src/main/scala/temple/DSL/semantics/ServiceRenamer.scala
```diff
@@ -0,0 +1,88 @@
...
+        case writable: Metadata.Writable        => writable
+      },
+    )
+
+  private def renameTargetBlocks(targets: Map[String, TargetBlock]): Map[String, TargetBlock] =
+    targets.view.mapValues(renameTargetBlock).toMap
+
+  private def renameServiceMetadata(metadata: Metadata.ServiceMetadata): Metadata.ServiceMetadata = metadata match {
+    // rename any services referenced in #uses
+    case Metadata.Uses(services) => Metadata.Uses(services.map(renaming))
```
jaylees14: This isn't too clear what's going on? You're using a map lookup as a function? Then what happens if `renaming` doesn't contain that key?

src/main/scala/temple/DSL/semantics/Validator.scala
```diff
@@ -1,44 +1,81 @@
...
 import temple.ast.{Metadata, _}
+import temple.builder.project.ProjectConfig
 
 import scala.collection.mutable
 import scala.reflect.{ClassTag, classTag}
 
-private class Validator private (templefile: Templefile) {
-  var errors: mutable.Set[String] = mutable.SortedSet()
+private class Validator(templefile: Templefile) {
+  var errors: mutable.Buffer[String] = mutable.Buffer()
```
jaylees14: Is the mutable prefix needed? Don't think it'd make much sense to have an `immutable.Buffer` 😆

src/main/scala/temple/DSL/semantics/Validator.scala
```diff
@@ -1,44 +1,81 @@
...
+  private var newServices: Map[String, ServiceBlock] = templefile.services
+
+  private val globalRenaming      = mutable.Map[String, String]()
+  def allGlobalNames: Set[String] = globalRenaming.valuesIterator.toSet ++ allStructsAndServices
+
+  private def validateAttributes(
+    attributes: Map[String, Attribute],
+    context: SemanticContext,
+  ): Map[String, Attribute] = {
+    foreachValueWithContext(attributes, context, validateAttribute)
```
jaylees14: There seems to be a lot of words here, but I'm still not quite sure what's going on? Is it really worth the function call or could we just put that single line here?

src/main/scala/temple/DSL/semantics/Validator.scala
```diff
@@ -1,44 +1,81 @@
...
+    attributes.foldLeft(attributes.mapFactory[String, Attribute]()) {
+      case (newAttributes, (attributeName, value)) =>
+        if (attributeName.headOption.exists(!_.isLower))
+          errors += context.errorMessage(
+            s"Invalid attribute name $attributeName, it must start with a lowercase letter,",
+          )
+        val newName = dodgeNames(
+          attributeName,
+          templefile.projectName,
+          (newAttributes.keys ++ attributes.keys).toSet - attributeName,
```
jaylees14: A comment explaining this might be nice

src/main/scala/temple/DSL/semantics/Validator.scala
```diff
@@ -158,6 +198,38 @@ object Validator {
...
+    validator.transformed
   }
+
+//  private type EntryValidator[T] = (String, T, SemanticContext) => Unit
+//
+//  @inline final private def foreachEntryWithContext[T](
+//    map: Map[String, T],
+//    context: SemanticContext,
+//    validate: EntryValidator[T],
+//  ): Unit = map.foreachEntry { case (name, t) => validate(name, t, context :+ name) }
```
jaylees14: Can we clean this up?

src/main/scala/temple/DSL/semantics/Validator.scala
```diff
@@ -1,44 +1,81 @@
...
+  private val globalRenaming      = mutable.Map[String, String]()
+  def allGlobalNames: Set[String] = globalRenaming.valuesIterator.toSet ++ allStructsAndServices
+
+  private def validateAttributes(
+    attributes: Map[String, Attribute],
+    context: SemanticContext,
+  ): Map[String, Attribute] = {
+    foreachValueWithContext(attributes, context, validateAttribute)
+
+    // This looks like a map, but it isn’t because the accumulator `newAttributes`’ names are accessed during iteration
```
jaylees14: Rather than saying what this should or shouldn't look like, maybe explain the map you're building up?

src/main/scala/temple/DSL/semantics/NameClashes.scala
```diff
@@ -0,0 +1,141 @@
...
+    }
+  }
+
+  val postgresValidator: NameValidator = NameValidator.fromIterable(postgresNames)
+
+  /** Repeatedly add the project name to the start of the string */
+  def nameSuggestions(name: String, project: String, decapitalize: Boolean = false): Iterator[String] =
+    iterate(name.capitalize) { project + _ }.map { if (decapitalize) StringUtils.decapitalize else identity }
+
+  def dodgeNames(name: String, project: String, takenNames: Set[String], decapitalize: Boolean = false)(
```
jaylees14: Hmm not sure I like `dodge`

jaylees14: Could this have a comment too? Maybe explaining what happens if the name is or isn't unique? The API usage isn't very clear here

src/main/scala/temple/DSL/semantics/NameClashes.scala
```diff
@@ -0,0 +1,141 @@
...
+    validators: NameValidator*,
```
jaylees14, line 141: Any reason why this was taken separately?

src/main/scala/temple/DSL/semantics/Validator.scala
```diff
@@ -1,44 +1,81 @@
...
+        newAttributes + (newName -> value)
     }
-    attributes.foreachEntry(context(validateAttribute))
   }
 
-  private def validateService(service: ServiceBlock)(context: SemanticContext): Unit = {
-    service.structs.foreachEntry(context(validateStruct))
-    validateAttributes(service.attributes)(context)
-    validateMetadata(service.metadata, service.attributes)(context)
+  private def renameBlock(name: String, block: TempleBlock[_]): Unit = {
```
jaylees14, line 63: A comment here, explaining the side effect please

jaylees14 (commented):
Changes looks mostly understandable. `Validate.scala` has now become pretty complicated, so I think we'd all appreciate a little more documentation in there, especially given that some functions have side effects.

Am I right in saying that there's the assumption that `globalRenaming` will contain exactly the map of everything in the project (that could be renamed) to its new name? If so that part seems pretty hard to follow, and I can't check that that is necessarily true.

src/main/scala/temple/DSL/semantics/NameClashes.scala
c-harding: What would you suggest I do instead?

src/main/scala/temple/DSL/semantics/ServiceRenamer.scala
c-harding: Yes, maps are functions. I could make it clearer if you’d like. If it doesn’t include the key it will throw a (maybe not particularly easy to read) error, I’ll update that to make it more readable.

src/main/scala/temple/DSL/semantics/Validator.scala
c-harding: Good point, was a mutable set so I’ll fix that

src/main/scala/temple/DSL/semantics/Validator.scala
c-harding: My concern was that there would be places where we forgot to append to the context if we did it directly, but I will try to see how it looks

src/main/scala/temple/DSL/semantics/Validator.scala
c-harding: Good point

src/main/scala/temple/DSL/semantics/NameClashes.scala
c-harding: Sure, will do. What would you suggest for the name?

src/main/scala/temple/DSL/semantics/NameClashes.scala
c-harding: It’s a varargs following an optional parameter, so I’d have to make `decapitalize` required to allow both.

src/main/scala/temple/DSL/semantics/Validator.scala
c-harding: Sure, coming soon

c-harding:
@jaylees14 documentation is on its way. globalRenaming does contain every service and struct, even if they’re not named. I’ll make that more explicit too.

src/main/scala/temple/DSL/semantics/NameClashes.scala
jaylees14: I think it’s fine as is, just made me think a little 😉

src/main/scala/temple/DSL/semantics/NameClashes.scala
jaylees14: Ah fair enough 😀

src/main/scala/temple/DSL/semantics/NameClashes.scala
jaylees14: Hmm, naming isn’t my strong point. `constructUniqueName`? @smithwjv is usually pretty good at this stuff

willjvsmith: What is it doing

jaylees14 (approved):
Awesome job @xsanda! Nice and readable, really good stuff

jaylees14 (approved):

src/main/scala/temple/DSL/semantics/NameClashes.scala
c-harding: @smithwjv taking a name for a service/struct/attribute, and changing it if necessary not to conflict with reserved names in any host language

willjvsmith: Looks like a `createValidName` or `constructValidName` to me

Lewiky (approved):
Cool LGTM

PR #237, opened 2020-04-05
[go-gen] Replace createdByAttribute with optional

willjvsmith:
* Replaces `CreatedByAttribute` with `Option[CreatedByAttribute]` in `ServiceRoot` and makes  it a simple case class with a boolean to indicate if used to filter enumeration
* Adds names to `ServiceRoot` arguments in tests

I realised that to implement readable and writable on the handlers, this will be switched on rather a lot, so I thought I'd tidy it up.

### DIFF
--- a/src/main/scala/temple/builder/ServerBuilder.scala
+++ b/src/main/scala/temple/builder/ServerBuilder.scala
@@ -30,25 +30,13 @@ object ServerBuilder {
       case ServiceLanguage.Go => GoLanguageConfig
     }
 
-    // TODO: This could be nicer
-    val createdBy: Option[CreatedByAttribute] = serviceBlock.lookupMetadata[ServiceEnumerable] match {
-      case Some(ServiceEnumerable(true)) =>
-        Some(
-          CreatedByAttribute(
-            languageConfig.createdByInputName,
-            languageConfig.createdByName,
-            filterEnumeration = true,
-          ),
+    val createdBy: Option[CreatedByAttribute] = serviceBlock.lookupMetadata[ServiceEnumerable].map {
+      case ServiceEnumerable(byThis) =>
+        CreatedByAttribute(
+          languageConfig.createdByInputName,
+          languageConfig.createdByName,
+          filterEnumeration = byThis,
         )
-      case Some(ServiceEnumerable(false)) =>
-        Some(
-          CreatedByAttribute(
-            languageConfig.createdByInputName,
-            languageConfig.createdByName,
-            filterEnumeration = false,
-          ),
-        )
-      case None => None
     }
 
     val idAttribute = IDAttribute("id")
--- a/src/main/scala/temple/builder/ServerBuilder.scala
+++ b/src/main/scala/temple/builder/ServerBuilder.scala
@@ -30,18 +30,25 @@ object ServerBuilder {
       case ServiceLanguage.Go => GoLanguageConfig
     }
 
-    val createdBy: CreatedByAttribute = serviceBlock.lookupMetadata[ServiceEnumerable] match {
+    // TODO: This could be nicer
+    val createdBy: Option[CreatedByAttribute] = serviceBlock.lookupMetadata[ServiceEnumerable] match {
       case Some(ServiceEnumerable(true)) =>
-        CreatedByAttribute.EnumerateByCreator(
-          languageConfig.createdByInputName,
-          languageConfig.createdByName,
+        Some(
+          CreatedByAttribute(
+            languageConfig.createdByInputName,
+            languageConfig.createdByName,
+            filterEnumeration = true,
+          ),
         )
       case Some(ServiceEnumerable(false)) =>
-        CreatedByAttribute.EnumerateByAll(
-          languageConfig.createdByInputName,
-          languageConfig.createdByName,
+        Some(
+          CreatedByAttribute(
+            languageConfig.createdByInputName,
+            languageConfig.createdByName,
+            filterEnumeration = false,
+          ),
         )
-      case None => CreatedByAttribute.None
+      case None => None
     }
 
     val idAttribute = IDAttribute("id")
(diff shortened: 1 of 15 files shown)


src/main/scala/temple/generate/server/CreatedByAttribute.scala
```diff
@@ -1,36 +1,11 @@
...
-}
+/**
+  * CreatedByAttribute encapsulates the temple defined attribute used to track which authenticated ID created a resource
+  *
+  * @param inputName the name of the variable used as input, e.g. "authID", present for improving semantics of generated
+  *                  code
+  * @param name the name of the attribute, e.g. "createdBy"
+  * @param filterEnumeration whether or not this attribute is used to filter the list operation
+  */
+case class CreatedByAttribute(inputName: String, name: String, filterEnumeration: Boolean)
```
jaylees14, line 11: Yeah nice, this hopefully removes quite a lot of `None => None` mappings :)

src/main/scala/temple/builder/ServerBuilder.scala
```diff
@@ -30,18 +30,25 @@ object ServerBuilder {
       case ServiceLanguage.Go => GoLanguageConfig
     }
 
-    val createdBy: CreatedByAttribute = serviceBlock.lookupMetadata[ServiceEnumerable] match {
+    // TODO: This could be nicer
+    val createdBy: Option[CreatedByAttribute] = serviceBlock.lookupMetadata[ServiceEnumerable] match {
```
jaylees14: There's actually quite a lot you can do here. If you replace the `match` with a `map`, you can remove the `None => None` case as well as the `Some` wrapping on the `CreatedByAttribute`. Then because both ServiceEnumerable cases are the same except for a boolean, we can just make the pattern match using a variable and use this to create the attribute.

All in all I think this looks like:

```scala
  val createdBy: Option[CreatedByAttribute] = serviceBlock.lookupMetadata[ServiceEnumerable].map {
      case ServiceEnumerable(filterByThis) =>
        CreatedByAttribute(
          languageConfig.createdByInputName,
          languageConfig.createdByName,
          filterEnumeration = filterByThis,
        )
    }
```

src/main/scala/temple/generate/server/go/service/GoServiceGenerator.scala
```diff
@@ -40,8 +40,14 @@ object GoServiceGenerator extends ServiceGenerator {
 
     // Whether or not this service has an auth block
     val hasAuthBlock = root.createdByAttribute match {
```
jaylees14: I think this is just `root.createdByAttribute.isEmpty`?

src/main/scala/temple/generate/server/go/service/GoServiceGenerator.scala
```diff
@@ -40,8 +40,14 @@ object GoServiceGenerator extends ServiceGenerator {
...
-      case CreatedByAttribute.None                                                         => true
-      case _: CreatedByAttribute.EnumerateByCreator | _: CreatedByAttribute.EnumerateByAll => false
+      case None                        => true
+      case Some(_: CreatedByAttribute) => false
+    }
+
+    // Whether or not this service is enumerating by creator
+    val enumeratingByCreator = root.createdByAttribute match {
```
jaylees14, line 48: You could even clean this up to 
```scala
val enumeratingByCreator = root.createdByAttribute.contains(CreatedByAttribute(_, _, filterEnumeration = true))
```

src/main/scala/temple/generate/server/go/service/dao/GoServiceDAOFunctionsGenerator.scala
```diff
@@ -29,8 +28,8 @@ object GoServiceDAOFunctionsGenerator {
       operation match {
         case List =>
           root.createdByAttribute match {
```
jaylees14: Again could use the `.contains` approach from above, but whatever you think is cleaner

src/main/scala/temple/generate/server/go/service/dao/GoServiceDAOFunctionsGenerator.scala
```diff
@@ -107,8 +106,8 @@ object GoServiceDAOFunctionsGenerator {
       .list(
         s"&${root.decapitalizedName}.${root.idAttribute.name.toUpperCase}",
         root.createdByAttribute match {
```
jaylees14: Because we're in optional world now we can just use `.map` which automatically does the `None => None` case for you :)

src/main/scala/temple/generate/server/go/service/dao/GoServiceDAOGenerator.scala
```diff
@@ -36,9 +36,9 @@ object GoServiceDAOGenerator {
   private[service] def generateDatastoreObjectStruct(root: ServiceRoot): String = {
     val idMap = ListMap(root.idAttribute.name.toUpperCase -> generateGoType(AttributeType.UUIDType))
     val createdByMap = root.createdByAttribute match {
```
jaylees14: Ditto here, `map` over `match`

src/main/scala/temple/generate/server/go/service/dao/GoServiceDAOInputStructsGenerator.scala
```diff
@@ -28,9 +28,9 @@ object GoServiceDAOInputStructsGenerator {
 
     // Note we use the createdBy input name, rather than name
     lazy val createdByMap = root.createdByAttribute match {
-      case CreatedByAttribute.None =>
+      case None =>
```
jaylees14: And here 😛

jaylees14 (commented):
Nice stuff. Just a couple of suggestions of how to tidy things up now we're dealing with optionals - we can take advantage of all the nice functions Scala provides 😄

jaylees14 (approved):
Nice one 💯

c-harding (approved):
Love it ❤️

PR #239, opened 2020-04-06
[e2e] Refactor end-to-end tests to use directory

jaylees14:
Ah. Quite big. You'll be wanting to just filter to `.scala` files 😄 

Writing E2E tests was getting annoying, and we were missing cases where we didn't fully check certain subdirectories etc. So: now we compare a directory in `resources` with the output files by building a map of both and then comparing. This means the test logic never has to change (unless you add a new templefile).

I found an issue with the way `FileMatchers` was working, it has been corrected to `actual shouldBe expected`. 

After this PR I'll go round and make sure all our tests do this, as it makes for some hard diffing 😢


src/e2e/resources/simple-temple-expected/auth-db/init.sql
```diff
@@ -0,0 +1,5 @@
+CREATE TABLE auth (
+  id UUID UNIQUE NOT NULL,
+  email TEXT UNIQUE NOT NULL,
+  password TEXT NOT NULL
+);
```
c-harding, line 5: File without trailing newline 😢

src/e2e/resources/simple-temple-expected/booking-db/init.sql
```diff
@@ -0,0 +1,3 @@
+CREATE TABLE booking (
+
+);
```
c-harding, line 3: Best table! Also, newline

src/e2e/resources/simple-temple-expected/grafana/provisioning/dashboards/simple-temple-test-user.json
```diff
@@ -882,4 +882,4 @@
   "title" : "SimpleTempleTestUser Dashboard",
   "uid" : "simple-temple-test-user",
   "version" : 1
-}
+}
```
c-harding, line 885: newline

src/e2e/scala/temple/E2ETestUtils.scala
```diff
@@ -0,0 +1,42 @@
...
+  /**
+    * Build a map of files present in the provided filepath
+    * @param path The root path to explore
+    * @return A map of files, relative to the provided path
+    */
+  def buildFileMap(path: Path): FileSystem.Files = {
+    var foundFiles: FileSystem.Files = Map.empty
+    Files.walkFileTree(
+      path,
+      new FileVisitor[Path] {
```
c-harding, line 22: such scotlandyard, very Tilo

c-harding (approved):
Feels like there’s a random mix of trailing newlines and not in the generated code. Not your fault, but GitHub makes it more obvious now

The scala looks beautifully OOP

willjvsmith (approved):
LGTM 🚀

PR #240, opened 2020-04-06
[Validator] Treat service blocks as struct blocks for ease of traversal

c-harding:
Just a quick one, should make dealing with both struct blocks and service blocks easier. Main result is that `structIterator` will return structs, whose metadata can be read, instead of simple maps of names to attributes


Lewiky (approved):
Cool cool cool

src/main/scala/temple/ast/OldServiceBlock.scala
```diff
@@ -0,0 +1,27 @@
+package temple.ast
+
+import temple.ast.Metadata.ServiceMetadata
+
+/** A service block, representing one microservice on its own isolated server */
+case class OldServiceBlock(
```
jaylees14: I can't see where this is used? Also old vs what?

src/main/scala/temple/ast/StructBlock.scala
```diff
@@ -2,8 +2,7 @@ package temple.ast
 
 import temple.ast.Metadata.StructMetadata
 
-/** A nested struct within a service, representing a second table/domain in the same datastore */
-case class StructBlock(
-  attributes: Map[String, Attribute],
-  metadata: Seq[StructMetadata] = Nil,
-) extends TempleBlock[StructMetadata]
+trait StructBlock[M >: StructMetadata <: Metadata] extends TempleBlock[M] {
```
jaylees14: Hmm I'm not sure about the naming `Struct` and `NestedStruct`. I feel like `Struct` needs to stay as is (to match Templefile syntax) and this becomes something `AbstractStruct` to make it clear it's a trait? Not sure about names though

jaylees14 (approved):
Couple of suggestions, but happy to go ahead if you don't agree

src/main/scala/temple/ast/OldServiceBlock.scala
c-harding: Sorry, this should have been deleted following a merge conflict

PR #241, opened 2020-04-06
[newline] Add newlines consistently to all generated files

jaylees14:
Hopefully this should fix all the newlines that were missing in #239 and since I was updating those tests, I fixed the parameter ordering to match the order of `shouldBe`


Lewiky (approved):

willjvsmith (approved):
LGTM 🚀

PR #242, opened 2020-04-06
[builder] Refactor project builder into distinct functions for each component

jaylees14:
It felt as though having one function in `ProjectBuilder` was getting quite messy, so I've separated it out into smaller functions. Some could do with a tidy up themselves, but I thought this was a good first step


willjvsmith (approved):
Look very sensible to me 🚀

Lewiky (approved):
Yeah I like this

src/main/scala/temple/builder/project/ProjectConfig.scala
```diff
@@ -9,12 +10,13 @@ object ProjectConfig {
...
-  val defaultLanguage: ServiceLanguage = ServiceLanguage.Go
-  val defaultDatabase: Database        = Database.Postgres
-  val defaultAuth: ServiceAuth         = ServiceAuth.Email
-  val authPort: Int                    = 1024
-  val authMetricPort: Int              = 1025
-  val serviceStartPort: Int            = 1026
+  val defaultLanguage: ServiceLanguage  = ServiceLanguage.Go
+  val defaultDatabase: Database         = Database.Postgres
+  val defaultAuth: ServiceAuth          = ServiceAuth.Email
+  val defaultPreparedType: PreparedType = PreparedType.DollarNumbers
```
c-harding: Shouldn’t this be a function of the language?

c-harding (approved):
lgtm, just one question

src/main/scala/temple/builder/project/ProjectConfig.scala
jaylees14: Good point, I'll fix

PR #243, opened 2020-04-06
[go-gen] Fix create handler to handle auth/project-auth

willjvsmith:
* Adds handling of services with and without auth, as well as projects with and without auth, to the create handler generation
  * Adds this handling to interservice-communication generation, i.e. `handler.go`
* Updates tests
  * Corrects the `simple-user` service output, since it doesn't have any `auth`

This handles `writable` behaviour and `auth/project-auth` behaviour, has been real fun to work out, finding other issues along the way, but we're here now! Some other stuff needs checking like the DAO for auth/non-auth - I've made a card for that.


c-harding (approved):
very nice

jaylees14 (approved):
All looks good to me :)

PR #244, opened 2020-04-06
[Validators] Add naming validators for Go, including functional validators

c-harding:
This also converts the NameValidator system to work with sets of values, instead of individual values. Go is more complex than postgres, because there are name clashes from our generated code, some of which are not constant strings. This also strips underscores etc, to ensure that valid strings aren’t converted to invalid strings in postgres when they are converted to snake case


src/main/scala/temple/DSL/semantics/NameClashes.scala
```diff
@@ -111,22 +188,30 @@ object NameClashes {
     "with",
   )
 
-  case class NameValidator private (private[NameClashes] val isValid: String => Boolean)
+  case class NameValidator private (private val validators: Set[String => Boolean]) {
+
+    def isValid(word: String): Boolean = validators.exists(validator => validator(NameValidator.normalize(word)))
+
+    // Combine two validators
+    def &(that: NameValidator): NameValidator = NameValidator(this.validators ++ that.validators)
```
jaylees14, line 240: You scala boy 😉

src/main/scala/temple/DSL/semantics/NameClashes.scala
```diff
@@ -111,22 +188,30 @@ object NameClashes {
...
+  }
 
   private[NameClashes] object NameValidator {
-    private val acceptEverything = NameValidator(_ => true)
+    private def normalize(string: String): String = string.toLowerCase.replaceAll("[^a-z0-9]", "")
```
jaylees14, line 244: So the idea of this is to remove anything like `-` or `_`?

jaylees14: I'm speaking hypothetically, but what happens if they had something like `foo-bar` and `fooBar` as two separate names - that can be dealt with?

jaylees14 (approved):
Nice stuff - would be good to see some tests, even if small 😄

src/main/scala/temple/DSL/semantics/NameClashes.scala
Lewiky, line 240: for real tho this api is sweet

Lewiky (approved):
Cool love it

PR #245, opened 2020-04-06
[go-gen] Add read handler generation

willjvsmith:
* Adds read handler generation
  * Handles with auth/project-auth
  * Handles `readable(by: this)` and `readable(by: all)`
* Adds some helper functions
* Renames `generateCheckAndReturnError` -> `genCheckAndReturnError`
* Updates tests

As usual, can mostly ignore the `.go` and `.go.snippet`. Since the complicated auth/project-auth and readable/writable logic was handled in the last PR,  generation has become much more pleasant 😄

I'm also thinking it would be a good idea to run the integration tests on the project builder test output.


src/main/scala/temple/generate/server/go/common/GoCommonGenerator.scala
```diff
@@ -139,11 +139,25 @@ object GoCommonGenerator {
     mkCode(
       "func",
       CodeWrap.parens(objectName, objectType),
-      CodeWrap.parens.prefix(methodName)(methodArgs),
+      CodeWrap.parens.prefix(methodName)(mkCode.list(methodArgs)),
```
c-harding: ….prefix(methodName).list(methodArgs) should work, from memory

src/main/scala/temple/generate/server/go/common/GoCommonGenerator.scala
```diff
@@ -139,11 +139,25 @@ object GoCommonGenerator {
...
       methodReturn,
       CodeWrap.curly.tabbed(methodBody),
     )
 
+  /** Generation function */
```
c-harding: “Generate a function” please, this parses as just a label that this function is used for generation.

src/main/scala/temple/generate/server/go/common/GoCommonGenerator.scala
```diff
@@ -139,11 +139,25 @@ object GoCommonGenerator {
...
+  private[go] def genFunc(
+    funcName: String,
+    funcArgs: Seq[String],
+    funcReturn: Option[String],
+    funcBody: String,
+  ): String =
+    mkCode(
+      "func",
+      CodeWrap.parens.prefix(funcName)(mkCode.list(funcArgs)),
```
c-harding: Ditto, prefix(funcName).list(…)

c-harding (approved):
lgtm

src/main/scala/temple/generate/server/go/service/main/GoServiceMainHandlersGenerator.scala
```diff
@@ -65,10 +65,31 @@ object GoServiceMainHandlersGenerator {
...
+  /** Generate one line http.Error call and return */
+  private[main] def generateOneLineHTTPErrorReturn(statusCodeEnum: GoHTTPStatus): String =
+    mkCode.lines(
+      generateOneLineHTTPError(statusCodeEnum),
+      genReturn(),
+    )
+
   /** Generate the block for extracting an AuthID from the request header */
-  private[main] def generateExtractAuthBlock(): String =
+  private[main] def generateExtractAuthBlock(usesVar: Boolean): String =
```
jaylees14, line 91: Could do with explaining what `usesVar` means?

src/main/scala/temple/generate/server/go/GoHTTPStatus.scala
```diff
@@ -2,5 +2,5 @@ package temple.generate.server.go
 
 object GoHTTPStatus extends Enumeration {
   type GoHTTPStatus = Value
-  val StatusBadRequest, StatusUnauthorized, StatusInternalServerError = Value
+  val StatusBadRequest, StatusUnauthorized, StatusNotFound, StatusInternalServerError = Value
```
Lewiky, line 5: Pedantic, but nothing about this seems to be go specific - could be moved up? But then again it doesn't actually matter

jaylees14 (approved):
Yeah nice - still feels a little weird we do the read twice, but that's something we can optimise later :)

Lewiky (approved):
LGTM

src/main/scala/temple/generate/server/go/GoHTTPStatus.scala
willjvsmith: Is Go specific because it matches the syntax of the Go `http` package - though if we were supporting different languages I agree we would build a global http enum, that then maps into each language's enum, but ain't nobody got time for dat.

PR #248, opened 2020-04-07
[Parser] Remove underscores from Templefile variable names

c-harding:
(and basically any other special characters too, alphanumeric-only for maximum compatibility)

This also improves the testing of the parser, boosting fluency and checking the error messages when parsing goes wrong


jaylees14 (approved):
`it shouldBe approved`

Lewiky (approved):
LGTM

PR #249, opened 2020-04-07
[grafana] Add support for structs in Grafana

jaylees14:
Add support for showing structs in a Grafana dashboard.

This involves a few changes:
- The metric title becomes `Create <struct> Requests` rather than `Create <service> Requests`
- The parameter for each main service endpoint is `create`, `read`, `update` or `delete`, whereas for structs this becomes `create_<struct>`, `read_<struct>` etc.

The test changes are just additive, only updating the cases where structs are used.

Also cleaned up the metric builder function to make it a little more readable.


c-harding (approved):
very nice

Lewiky (approved):
LGTM bud-e

PR #250, opened 2020-04-07
[Validator] Validate readable/writable by this has an auth block

c-harding:
Validate that any templefile containing the metadata `#readable(by: this)`/`#writable(by: this)` uses auth, because otherwise there’s no notion of createdBy, which `by:this` relies on.

A couple of tests which this broke have been updated


Lewiky (approved):
Nice one

src/main/scala/temple/DSL/semantics/Validator.scala
```diff
@@ -119,9 +119,13 @@ private class Validator private (templefile: Templefile) {
       case _: Metadata.ServiceLanguage => assertUnique[Metadata.ServiceLanguage]()
       case _: Metadata.Database        => assertUnique[Metadata.Database]()
       case _: Metadata.ServiceAuth     => assertUnique[Metadata.ServiceAuth]()
-      case _: Metadata.Readable        => assertUnique[Metadata.Readable]()
+      case Metadata.Readable.This if !templefile.usesAuth =>
+        errors += context.errorMessage(s"#readable(this) requires at least one service to have #auth")
+      case _: Metadata.Readable => assertUnique[Metadata.Readable]()
       case Metadata.Writable.All if metadata contains Metadata.Readable.This =>
         errors += context.errorMessage(s"#writable(all) is not compatible with #readable(this)")
```
willjvsmith, line 126: Hadn't thought of this, but makes sense since you need to know ID to write

willjvsmith (approved):
LGTM 🚀

jaylees14 (approved):
Nice one

PR #251, opened 2020-04-07
[go-gen] Add delete handler generation

willjvsmith:
* Adds generation of delete handlers
  * Should handle readable/writable/auth/project-auth cases
* Improves checkAuthorization generation
  * Removes it from being generated for services with an auth block, as the ID is a one-to-one mapping with AuthID, so the incoming AuthID can be directly compared with the resource ID to check authorization

Standard stuff


src/main/scala/temple/generate/server/go/service/main/GoServiceMainHandlersGenerator.scala
```diff
@@ -117,21 +117,28 @@ object GoServiceMainHandlersGenerator {
...
+          genFunctionCall("checkAuthorization", "env", s"${root.decapitalizedName}ID", "auth"),
+          "authorized",
+          "err",
         ),
-      ),
-      genIf("!authorized", generateHTTPErrorReturn(StatusUnauthorized, "Unauthorized")),
-    )
+        genIfErr(
+          genSwitchReturn(
+            "err.(type)",
```
c-harding, line 134: wat

c-harding (approved):
LGTM

src/main/scala/temple/generate/server/go/service/main/GoServiceMainHandlersGenerator.scala
willjvsmith: nice syntax right

jaylees14 (approved):

PR #252, opened 2020-04-07
[dsl] Support #metrics in Templefile

jaylees14:
Support `#metrics(prometheus)` at the project level.

Ultimately we'd like to be able to specify the dashboard used as well, but since we don't support multiple argument metadata at the moment, it isn't worth it.


willjvsmith (approved):
Nice and clean 🚀

Lewiky (approved):
LGTM

PR #254, opened 2020-04-07
[config] Map into config.json

jaylees14:
This generates `config.json` for each service. 

Couple of things were found whilst doing this:
- Computing which services can be accessed also needs to take into account any structs within that service
- Go looks service keys up using `CamelCase` names, but then the URL needs `kebab-case` - are we okay with these being different?
- I don't like the user of `.get`, but we're pretty certain it's going to be in there - what should I do instead? Throw?
- None of our examples in ProjectBuilder or E2E test actually include a foreign key.
  - Rather than add one and have to deal with Go changes I just created a new test

### DIFF
--- a/src/e2e/resources/simple-temple-expected/booking/config.json
+++ b/src/e2e/resources/simple-temple-expected/booking/config.json
@@ -1,0 +1,13 @@
+{
+  "user" : "postgres",
+  "dbName" : "postgres",
+  "host" : "booking-db",
+  "sslMode" : "disable",
+  "services" : {
+    
+  },
+  "ports" : {
+    "service" : 1028,
+    "prometheus" : 1029
+  }
+}
--- a/src/main/scala/temple/builder/project/ProjectBuilder.scala
+++ b/src/main/scala/temple/builder/project/ProjectBuilder.scala
@@ -15,6 +15,7 @@ import temple.generate.metrics.grafana.ast.Datasource
 import temple.generate.metrics.grafana.{GrafanaDashboardConfigGenerator, GrafanaDashboardGenerator, GrafanaDatasourceConfigGenerator}
 import temple.generate.metrics.prometheus.PrometheusConfigGenerator
 import temple.generate.metrics.prometheus.ast.PrometheusJob
+import temple.generate.server.config.ServerConfigGenerator
 import temple.generate.server.go.auth.GoAuthServiceGenerator
 import temple.generate.server.go.service.GoServiceGenerator
 import temple.generate.target.openapi.OpenAPIGenerator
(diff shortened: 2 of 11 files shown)


src/main/scala/temple/builder/project/ProjectBuilder.scala
```diff
@@ -114,9 +115,18 @@ object ProjectBuilder {
       case (name, service, port) =>
         val serviceRoot =
           ServerBuilder.buildServiceRoot(name, service, port.service, endpoints(service), detail, usesAuth)
-        service.lookupMetadata[ServiceLanguage].getOrElse(ProjectConfig.defaultLanguage) match {
+        val serverFiles = service.lookupMetadata[ServiceLanguage].getOrElse(ProjectConfig.defaultLanguage) match {
           case ServiceLanguage.Go => GoServiceGenerator.generate(serviceRoot)
         }
+
+        val serviceComms = serviceRoot.comms.map { service =>
+          val (_, _, ports) = templefile.providedServicesWithPorts.find { _._1 == service }.get
```
c-harding, line 123: I see, maybe worth extracting this to a function in `Templefile`? But it’s only unsafe if the service isn’t really in the templefile

c-harding (approved):
looks good to me. is the kebabbing of file names going to be problematic?

src/main/scala/temple/builder/project/ProjectBuilder.scala
```diff
@@ -114,9 +115,18 @@ object ProjectBuilder {
...
+          service -> s"http://${kebabCase(service)}:${ports.service}"
+        }.toMap
+
+        val configFileContents =
+          ServerConfigGenerator.generate(serviceRoot.kebabName, serviceRoot.datastore, serviceComms, port)
+        serverFiles + (File(serviceRoot.kebabName, "config.json") -> configFileContents)
```
Lewiky:
```suggestion
        val configFileContents =
          ServerConfigGenerator.generate(serviceRoot.kebabName, serviceRoot.datastore, serviceComms, port)
          
        serverFiles + (File(serviceRoot.kebabName, "config.json") -> configFileContents)
```

Lewiky (approved):
LGTM

src/e2e/resources/simple-temple-expected/booking/config.json
```diff
@@ -0,0 +1,13 @@
+{
+  "user" : "postgres",
+  "dbName" : "postgres",
+  "host" : "booking-db",
+  "sslMode" : "disable",
+  "services" : {
+    
+  },
```
willjvsmith, line 8: I guess it doesn't matter, empty block

PR #255, opened 2020-04-07
[go-gen] Fix handling of @serverSet attributes in DAO generation

willjvsmith:
* Fixes handling of `@serverSet` attributes in the DAO
  * Previously `matchedOn` for example was provided in the SQL query as `NOW()`, but we don't know what this value will always be
  * The client attributes, which omit `@serverSet` attributes are passed to the DAO Create and Update input structs. Since the `@serverSet` attributes are omitted, they take their zero value, but can be overruled in the hooks!

This should make the go generation ready for use! Aside from some (large) edge cases (see cards).


c-harding (approved):
_molto bene_

Lewiky (approved):
What a boi

jaylees14 (approved):
LGTM

PR #256, opened 2020-04-07
[Kube-Gen] Deploy.sh 

Lewiky:
Adds generation for the deploy.sh file we used to run everything through kube. Pretty constant apart from adding one extra line per service, and a conditional section added when metrics are used.


src/main/resources/shell/deploy.sh.header.snippet
```diff
@@ -0,0 +1,20 @@
...
+
+minikube start --vm-driver=virtualbox
+
+echo $GREEN
+
+kubectl create secret docker-registry regcred --docker-server=$REG_URL --docker-username=$REG_USERNAME --docker-password=$REG_PASSWORD --docker-email=$REG_EMAIL
+
+echo
+echo Creating ConfigMaps...
+echo
```
c-harding: newline plz

src/main/resources/shell/deploy.sh.metrics.snippet
```diff
@@ -0,0 +1,12 @@
...
+
+# Grafana
+kubectl create configmap grafana-datasource-config --from-file "$BASEDIR/grafana/provisioning/datasources/datasource.yml" -o=yaml
+
+for file in "$BASEDIR/grafana/provisioning/dashboards/"*
+do
+  filename=$(basename $file) # Get everything after the final /
+  file=$(echo $filename | cut -f 1 -d '.') # Get everything before the . (filename without extension)
+  kubectl create configmap "grafana-$file-config" --from-file "$BASEDIR/grafana/provisioning/dashboards/$filename" -o=yaml
+done
```
c-harding: newline plz

src/test/scala/temple/TestUtils.scala
```diff
@@ -24,7 +24,10 @@ object TestUtils {
           FileVisitResult.CONTINUE
 
         override def visitFile(file: Path, attrs: BasicFileAttributes): FileVisitResult = {
-          foundFiles += File(path.relativize(file).getParent.toString, file.getFileName.toString) -> FileUtils.readFile(
+          foundFiles += File(
+            Option(path.relativize(file).getParent).map(_.toString).getOrElse(""),
```
c-harding, line 28: When does this return `null`?

c-harding (approved):
nice

src/test/scala/temple/TestUtils.scala
Lewiky: When the `file` you're relativizing to is in the same directory as `path`,  `getParent` is `null`

jaylees14: Nice that 😉

jaylees14 (approved):
Looks good, think you need to merge

PR #260, opened 2020-04-07
[it] Add integration test against simple.temple

jaylees14:
Turns out `simple.temple` doesn't compile 😫... so let's make it an integration test.
I couldn't be bothered to build a `Templefile` that matches `simple.temple` so I'm using the normal pipeline of raw string → `Templefile`. This means if you make change to `simple.temple`, it should now break the integration tests.

Here are the current errors - I've commented the actual test out so it's easier to see what the fixes were: 

Screenshot 2020-04-07 at 19 01 00

### DIFF
--- a/src/it/scala/temple/generate/server/GoGeneratorIntegrationTest.scala
+++ b/src/it/scala/temple/generate/server/GoGeneratorIntegrationTest.scala
@@ -1,9 +1,15 @@
 package temple.generate.server
 
 import org.scalatest.{BeforeAndAfter, Matchers}
+import temple.DSL.DSLProcessor
+import temple.DSL.semantics.Analyzer
+import temple.builder.project.ProjectBuilder
 import temple.containers.GolangSpec
+import temple.detail.LanguageDetail.GoLanguageDetail
 import temple.generate.FileSystem._
 import temple.generate.server.go.service.GoServiceGenerator
+import temple.utils.{FileUtils, StringUtils}
+import temple.utils.MonadUtils.FromEither
 
 class GoGeneratorIntegrationTest extends GolangSpec with Matchers with BeforeAndAfter {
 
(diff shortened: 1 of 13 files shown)


src/it/scala/temple/generate/server/GoGeneratorIntegrationTest.scala
```diff
@@ -85,4 +91,26 @@ class GoGeneratorIntegrationTest extends GolangSpec with Matchers with BeforeAnd
...
+
+    analyzedTemplefile.allServices.keys.map(StringUtils.kebabCase).foreach { service =>
+      val filesInService = project.files.filter { case (file, _) => file.folder.startsWith(service) }
+      val validationErrors = validateAll(
+        filesInService,
+        entryFile = File(service, s"$service.go"),
+      )
+
+    // TODO: Uncomment me, and fix the bugs!
+//      validationErrors shouldBe empty
```
c-harding: ooh, exciting

c-harding (approved):
nice test, 100% pass rate

src/it/scala/temple/generate/server/GoGeneratorIntegrationTest.scala
willjvsmith: dealing with it now

willjvsmith (approved):

src/main/scala/temple/generate/server/go/service/main/GoServiceMainHandlersGenerator.scala
```diff
@@ -19,12 +19,16 @@ import scala.Option.when
...
 object GoServiceMainHandlersGenerator {
 
-  private def generateResponseMapFormat(attributeType: AttributeType): String =
-    // Must add formatting to attributes with datetime type
-    attributeType match {
-      case DateTimeType => s".${genFunctionCall("Format", "time.RFC3339")}"
+  private def generateResponseMapFormat(attributeType: AttributeType): String = {
+    // Must add formatting to attributes with date, time or datetime type
+    val layout = attributeType match {
+      case DateType     => doubleQuote("2020-12-31")
```
jaylees14: @smithwjv See my gist (https://gist.github.com/jaylees14/e86b262ca4e0b8185cb5357c78fc01f3) for the correct format here. You need to use the reference time: https://gobyexample.com/time-formatting-parsing

Lewiky (approved):
LGTM

src/main/scala/temple/generate/server/go/service/main/GoServiceMainHandlersGenerator.scala
willjvsmith: Was wondering how they told the difference between days and months aha

PR #261, opened 2020-04-07
[hooks] Remove arguments if service has no attributes

jaylees14:
Turns out that we weren't compiling the entire project - so files that weren't getting used (such as `Hooks.go` weren't being checked. I've updated the Dockerfile, so they now are - and fixed some issues with hooks generation.

This mainly comes down to dealing with no attributes correctly - and differentiating between the two types of list endpoints.

### DIFF
--- a/docker/deploy.sh
+++ b/docker/deploy.sh
@@ -2,7 +2,7 @@
 
 BASE_REPO_URL=jaylees/templeeight
 SERVICE="shell2http"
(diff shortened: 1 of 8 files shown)


willjvsmith (approved):
LGTM 🚀

Lewiky (approved):
LGTM

Lewiky (approved):
LGTM

PR #262, opened 2020-04-07
[docker-gen] Remove go.sum from Dockerfile

jaylees14:
Turns out we don't generate `go.sum`, so we can't copy it into the Docker container....

It gets automatically generated in the container, so safe to remove here


willjvsmith (approved):

c-harding (approved):
-Σ

PR #263, opened 2020-04-08
[local-registry] Add deployment yaml files for local registry service

jaylees14:
Adds yaml files for pushing to a local repository. These were taken from https://gist.github.com/coco98/b750b3debc6d517308596c248daf3bb1, but separated into files because that's the way we roll

### DIFF
--- a/src/main/scala/temple/generate/kube/KubernetesGenerator.scala
+++ b/src/main/scala/temple/generate/kube/KubernetesGenerator.scala
@@ -3,7 +3,7 @@ package temple.generate.kube
 import io.circe.generic.auto._
 import io.circe.syntax._
 import io.circe.yaml.Printer
-import temple.generate.FileSystem._
+import temple.generate.FileSystem.{File, _}
 import temple.generate.kube.ast.OrchestrationType._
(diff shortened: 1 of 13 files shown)


willjvsmith (approved):

src/main/scala/temple/generate/kube/KubernetesGenerator.scala
```diff
@@ -190,6 +190,14 @@ object KubernetesGenerator {
...
   )
 
+  private val deployFiles: Files = Map(
+    File("kube/deploy", "deploy-daemon-set.yaml") -> FileUtils.readResources("kube/deploy/deploy-daemon-set.yaml"),
+    File("kube/deploy", "deploy-replication-controller.yaml") -> FileUtils.readResources(
+      "kube/deploy/deploy-replication-controller.yaml",
+    ),
+    File("kube/deploy", "deploy-service.yaml") -> FileUtils.readResources("kube/deploy/deploy-service.yaml"),
+  )
+
```
c-harding, line 200: So everything is fixed? Nice and simple, we’ve built a folder duplicator 😝

c-harding (approved):
looks good

PR #264, opened 2020-04-08
[deploy] Deploy registry service

jaylees14:
Spins up the registry before all other containers, then invokes the push-image script once ready.

Also fixes a bug where the kong configuration was renamed and includes the url `localhost:5000` in the image name.

Test by running `temple generate` on `simple.temple` - then

- To your `/etc/hosts/` add a mapping to `registry.me` from localhost:
```
127.0.0.1|localhost registry.me
```
- update `push-image.sh` to use `registry.me:5000` instead of `localhost:5000`
- run `source deploy.sh`


c-harding (approved):
Reading this, I forgot it was even a Scala project

Lewiky (approved):
Love it

PR #265, opened 2020-04-08
[go-gen] Fix declare case and blob type encoding

willjvsmith:
* Fixes a case where a service is in a project without auth and has client attributes, variable wasn getting assigned before being declared
* Fixes encoding of blob types to base64
  * Includes import generation

I thought I would try compile the temple-user service from the project builder tests, since it's our only set of testfiles that doesn't have auth in the project. Compiles now :)


Lewiky (approved):
Very good

jaylees14 (approved):
Nice

PR #266, opened 2020-04-08
[CreatedBy] Add implicit CreatedBy Attribute where required

Lewiky:
Same as with the implicit `IDAttribute`, but when required adds the `CreatedBy` attribute. 90% of the changes here are just rewriting Attribute to be `AbstractAttribute` in types and changing imports since I had to refactor `Attribute` to extend `AbstractAttribute` to allow for pattern matching properly


src/e2e/resources/simple-temple-expected/booking/dao/dao.go
```diff
@@ -108,7 +108,7 @@ func (dao *DAO) ReadBooking(input ReadBookingInput) (*Booking, error) {
 
 // UpdateBooking updates the booking in the datastore for a given ID, returning the newly updated booking
 func (dao *DAO) UpdateBooking(input UpdateBookingInput) (*Booking, error) {
-	row := executeQueryWithRowResponse(dao.DB, "UPDATE booking SET WHERE id = $1 RETURNING id;", input.ID)
+	row := executeQueryWithRowResponse(dao.DB, "UPDATE booking SET WHERE id = $1 RETURNING id, createdBy;", input.ID)
```
jaylees14, line 111: This is invalid, but being fixed by @xsanda right?

src/main/scala/temple/builder/DatabaseBuilder.scala
```diff
@@ -53,14 +54,15 @@ object DatabaseBuilder {
...
     endpoints: Set[CRUD],
     readable: Metadata.Readable = Metadata.Readable.This,
     selectionAttribute: String = "id",
   ): ListMap[CRUD, Statement] = {
-    val tableName        = StringUtils.snakeCase(serviceName)
-    val columns          = attributes.keys.map(Column).toSeq
-    val columnsWithoutID = attributes.filter { case (_, attr) => attr != IDAttribute }.keys.map(Column).toSeq
+    val tableName = StringUtils.snakeCase(serviceName)
+    val columns   = attributes.keys.map(Column).toSeq
+    val columnsWithoutID =
```
willjvsmith: Might wanna rename

jaylees14 (approved):
Nice

src/e2e/resources/simple-temple-expected/booking/dao/dao.go
willjvsmith, line 111: @xsanda is fixing so this never actually gets generated, since the UPDATE endpoint will be omitted higher up in the chain

willjvsmith (approved):
LGTM 🚀

PR #267, opened 2020-04-08
[Validator] Omit the update endpoint when there’s nothing to update

c-harding:
Wow, that’s killed a surprising amount of generated code. Side effects are slightly easier updating of the test material in ProjectBuilderTest, but extracting the folder writing to a function and documenting how to update it.


src/main/scala/temple/DSL/semantics/Validator.scala
```diff
@@ -112,9 +112,33 @@ private class Validator private (templefile: Templefile) {
...
+      var foundOmit = false
+      val newMetadata = metadata map {
+          case Metadata.Omit(endpoints) =>
+            foundOmit = true
+            Metadata.Omit(endpoints + Metadata.Endpoint.Update)
+          case m => m
+        }
+      if (foundOmit) newMetadata
+      else newMetadata :+ Metadata.Omit(Set(Metadata.Endpoint.Update))
+    } else metadata
```
c-harding, line 136: This whole function seems a little odd in its control follow, anyone got a better suggestion of a way to lay it out?

Lewiky (approved):
Looks good - see what you mean with the noUpdates stuff being a little weird but not sure I have anything better

src/main/scala/temple/DSL/semantics/Validator.scala
jaylees14: See what you mean, could you add a comment to explain why you're doing what you're doing

jaylees14 (approved):

PR #268, opened 2020-04-08
[docker-compose gen] Allow docker-compose and kube in DSL

Lewiky:
Allows `#provider(kubernetes)` and `#provider(docker-compose)` in Templefile

### DIFF
--- a/src/main/scala/temple/ast/Metadata.scala
+++ b/src/main/scala/temple/ast/Metadata.scala
@@ -34,12 +34,14 @@ object Metadata {
     case object Go extends ServiceLanguage("Go", "golang")
   }
 
-  sealed abstract class Provider private (name: String) extends EnumEntry(name) with ProjectMetadata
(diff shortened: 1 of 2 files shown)


jaylees14 (approved):

willjvsmith (approved):

c-harding (changes requested):
this won‘t work

src/main/scala/temple/ast/Metadata.scala
```diff
@@ -38,7 +38,8 @@ object Metadata {
 
   object Provider extends Enum[Provider] {
     val values: IndexedSeq[Provider] = findValues
-    case object AWS extends Provider("aws")
+    case object Kubernetes    extends Provider("kubernetes")
+    case object DockerCompose extends Provider("docker-compose")
```
c-harding: This is not a valid string in the templefile. try `dockerCompose` instead

c-harding (approved):
love it

PR #272, opened 2020-04-08
[Docker-Compose-Gen] Only generate kubernetes or docker compose when …

Lewiky:
…defined in templefile


jaylees14 (approved):
LOVE the negative diff <3

willjvsmith (approved):
LGTM 🚀

PR #274, opened 2020-04-08
[go-gen] Add generation of parsing time attributes in requests

willjvsmith:
* Adds generation of parsing blocks for attributes of type `date`, `time` or `datetime`
  * Request struct now makes these a string type, and are parsed before passing into DAO input struct
  * Cannot do inline because of error handling
* Optimises imports
* Updates tests

This bug was discovered with `temple test`, @jaylees14 tested this fix.


src/main/scala/temple/generate/server/go/service/main/GoServiceMainHandlersGenerator.scala
```diff
@@ -226,6 +226,64 @@ object GoServiceMainHandlersGenerator {
...
 
+  private def generateParseTimeBlock(name: String, attributeType: AttributeType): String =
+    mkCode.lines(
+      genDeclareAndAssign(
+        genMethodCall(
+          "time",
+          "Parse",
+          attributeType match {
+            case DateType     => doubleQuote("2006-01-02")
+            case TimeType     => doubleQuote("15:04:05.999999999")
```
Lewiky, line 237: What do these hard coded values mean?

jaylees14: They're layouts according to the reference time of 
`January 2nd at 15:04:05 2006 in timezone +07:00` (1-2-3-4-5-6-7)

https://gobyexample.com/time-formatting-parsing

jaylees14 (approved):
LGTM

src/main/scala/temple/generate/server/go/service/main/GoServiceMainHandlersGenerator.scala
willjvsmith: Dumb right

Lewiky (approved):
sure

PR #275, opened 2020-04-08
[docker-compose] Add support for services, databases and metrics

jaylees14:
Adds support for services, databases and metrics. Just need to do kong, but that's mostly just fixed strings.

Will also add integration testing & E2E testing in a separate PR.


src/test/scala/temple/generate/orchestration/dockercompose/DockerComposeGeneratorTest.scala
```diff
@@ -7,8 +7,15 @@ class DockerComposeGeneratorTest extends FlatSpec with Matchers {
 
   behavior of "DockerComposeGenerator"
 
-  it should "generate an empty file" in {
-    val generated = DockerComposeGenerator.generate("example", UnitTestData.basicOrchestrationRootWithMetrics)
-    generated.head._2 shouldBe ""
+  it should "generate a docker-compose file for a service without metrics" in {
+    val (_, fileContents) =
```
willjvsmith, line 11: Just curious, but could this be `_ -> fileContents` instead?

willjvsmith (approved):
Docker compose LGTM 🚀 don't know much about how the `circe` stuff works

src/test/scala/temple/generate/orchestration/dockercompose/DockerComposeGeneratorTest.scala
c-harding: It could indeed. It might look a little odd, but so does so much of Scala

src/test/scala/temple/generate/orchestration/dockercompose/DockerComposeGeneratorTest.scala
```diff
@@ -7,8 +7,15 @@ class DockerComposeGeneratorTest extends FlatSpec with Matchers {
...
+      DockerComposeGenerator.generate("example", UnitTestData.basicOrchestrationRootWithoutMetrics).head
```
c-harding, line 12: Is this head safe?

c-harding (approved):
`circe` LGTM 🚀 don't know much about how the docker-compose stuff works

src/test/scala/temple/generate/orchestration/dockercompose/DockerComposeGeneratorTest.scala
jaylees14: Yes, always one is returned 😄

PR #276, opened 2020-04-08
[docker-compose] Generate kong scripts

jaylees14:
- Adds kong service configuration, as per spec-golang
- Fixes an issue with networks where they should be objects, not a list

- Ran these files against `docker-compose config`, which validated they were correct (maybe I should add that to an integration test...)


Lewiky (approved):
LGTM

willjvsmith (approved):
LGTM 🚀

PR #277, opened 2020-04-08
Auth docker compose

jaylees14:
When outputting an auth service for docker compose we need to use a special script that waits for Kong to be available before starting the service. See https://github.com/TempleEight/spec-golang/pull/20 for a full justification.

This adds support for that, as well as some unit tests.

### DIFF
--- a/src/main/scala/temple/builder/DockerfileBuilder.scala
+++ b/src/main/scala/temple/builder/DockerfileBuilder.scala
@@ -1,5 +1,6 @@
 package temple.builder
 
+import temple.ast.Metadata.Provider
 import temple.ast.{AbstractServiceBlock, Metadata}
 import temple.builder.project.ProjectConfig
 import temple.generate.docker.ast.Statement._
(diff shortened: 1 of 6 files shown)


Lewiky (approved):
LGTM my dude

src/main/scala/temple/builder/DockerfileBuilder.scala
```diff
@@ -8,10 +9,29 @@ import temple.generate.docker.ast.{DockerfileRoot, Statement}
...
 
+    // If using DockerCompose and this is the auth service, change the exec
+    val execStatements: Option[Seq[Statement]] = service match {
+      case AbstractServiceBlock.AuthServiceBlock =>
+        provider.collect {
+          case Provider.DockerCompose =>
+            Seq(
+              Run("apk", Seq("add", "curl")),
+              Run("chmod", Seq("+x", "wait-for-kong.sh")),
+              Cmd("./wait-for-kong.sh", Seq("kong:8001", "--", "./auth")),
```
willjvsmith, line 29: Is Kong always on 8001, or did we pick that

willjvsmith (approved):
LGTM 🚀

src/main/scala/temple/builder/DockerfileBuilder.scala
jaylees14: Always 😄

PR #280, opened 2020-04-09
[Docker-Compose-Gen] Deploy script generation DC

Lewiky:
Generates a `deploy.sh` for docker-compose, luckily this is static so we can just load it from a file :)


c-harding (approved):
nice

jaylees14 (approved):
LGTM

PR #281, opened 2020-04-09
[go-gen] Fix validation annotation

willjvsmith:
Annotations expected primitives, when they were checking pointers

PR #282, opened 2020-04-09
[Temple-test] Fix precision of numbers generated

c-harding:
On top of #281 

Generates numbers in the correct range based on the precision specified in the database.

Also adds methods to Attribute to determine whether they are in the request or the response, to simplify logic calculating that project wide


jaylees14 (approved):

c-harding:
Closed in favour of #283

PR #283, opened 2020-04-09
[Tester] Add precision checks to temple-testing numerical fields

c-harding:
Retrying the essence of #282, basically prevents doubles too large for a float being submitted to float fields by `temple test`
#


willjvsmith (approved):

willjvsmith:
I presume you tested this with `temple test` and it came up with the validator errors

jaylees14 (approved):

PR #285, opened 2020-04-09
[AST] Add attribute method for determining whether in request/response

c-harding:
Rather than constantly checking for `attribute.accessAnnotation.contains` everywhere, add methods on `attribute` for determining whether it should be present in a request to or a response from the server.

Codecov down due only to reduction in LOC.


jaylees14 (approved):
Nice

Lewiky (approved):
Nice one, v sensible

PR #286, opened 2020-04-09
[project-test] Tidy output from tester

jaylees14:
- Silence output from docker compose
- Remove env-var requirement from docker compose (might be able to remove from Kubernetes, not sure yet)
- Continue with tests if any fail - but still outputting a non-zero exit code at the end
- Add simple-dc.temple which is identical to simple.temple except for docker compose


willjvsmith (approved):
LGTM 🚀

Lewiky (approved):
Love

PR #288, opened 2020-04-09
[Validator] Safe test number sizing

c-harding:
Check that the provided minimum IntType bound is less than the maximum allowed for the given precision, and that the maximum IntType bound is greater than the minimum allowed.

This also shares the new methods for determining these bounds with the tester in its generation of random values.


jaylees14 (approved):

Lewiky (approved):
Cooooool

PR #287, opened 2020-04-09
[go-gen] Add JSON annotations to req & resp structs

willjvsmith:
* Adds JSON annotations to require keys be decapitalized

PR #291, opened 2020-04-09
[project-test] Remove explicit ID check and make createdBy @server

jaylees14:
- `createdBy` is now `@server` rather than `@serverSet`, since it's never returned
- Fixed some fail statements so they're easier to debug
- Remove explicit addition of `ID` field, since this is added by the templefile validator
- Remove `@unique` from boolean field in `simple-dc.temple`, since this makes it hard to test > 2 values
- Remove capitalisation of field names after #287


willjvsmith (approved):
LGTM 🚀

Lewiky (approved):
LGTM

Issue #290, opened 2020-04-09
Temple test freezes before configure kong stage when Templefile doesn't use Auth


Lewiky:
Minimal example:

```
Foreign: project{
  #language(go);
  #database(postgres);
  #provider(dockerCompose);
}

Foo: service {
  field: int;
  bar: Bar;
}

Bar: service {
  test: int;
}
```

Running `temple test` on this Templefile freezes after the docker-compose stage - the `configure-kong.sh` step never happens.

Adding an `#auth` block to one of the services fixes it

PR #295, opened 2020-04-09
[infra] Update temple test to allow templefiles without auth services

Lewiky:
The way temple-test had been written it was always assumed an auth service existed, so would try and get an auth token before starting, now we skip that logic if there's no auth service generated

### DIFF
--- a/src/main/scala/temple/test/ProjectTester.scala
+++ b/src/main/scala/temple/test/ProjectTester.scala
@@ -27,23 +27,21 @@ object ProjectTester {
         s"KONG_ADMIN=${config.kongAdminURL} KONG_ENTRY=${config.baseIP} sh $generatedPath/kong/configure-kong.sh",
       )
 
-      if (usesAuth) {
-        try {
-          configuredKong =
-            if (usesAuth)
-              Http("http://localhost:8000/api/auth/login")
-                .method("POST")
-                .asString
-                .code == HttpURLConnection.HTTP_BAD_REQUEST
-            // TODO: this might not be true, needs investigating
-            else true
-          if (!configuredKong) {
-            println("Kong wasn't configured correctly - trying again")
-            exec("sleep 5")
-          }
-        } catch {
-          case _: IOException => // Keep trying
+      try {
+        configuredKong =
+          if (usesAuth)
+            Http("http://localhost:8000/api/auth/login")
+              .method("POST")
+              .asString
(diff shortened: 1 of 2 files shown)


Lewiky:
Closes #290

jaylees14 (approved):
Nice catch 😄

src/main/scala/temple/test/ProjectTester.scala
```diff
@@ -27,23 +27,21 @@ object ProjectTester {
...
-        } catch {
-          case _: IOException => // Keep trying
+      try {
+        configuredKong =
+          if (usesAuth)
+            Http("http://localhost:8000/api/auth/login")
+              .method("POST")
+              .asString
+              .code == HttpURLConnection.HTTP_BAD_REQUEST
+          // TODO: this might not be true, needs investigating
```
c-harding, line 37: has this been investigated yet? If not, whose job is it?

src/main/scala/temple/test/ProjectTester.scala
```diff
@@ -27,23 +27,21 @@ object ProjectTester {
         s"KONG_ADMIN=${config.kongAdminURL} KONG_ENTRY=${config.baseIP} sh $generatedPath/kong/configure-kong.sh",
```
xsanda, line 27: Should this be in the while loop?

src/main/scala/temple/test/ProjectTester.scala
```diff
@@ -27,23 +27,21 @@ object ProjectTester {
...
-          case _: IOException => // Keep trying
+      try {
+        configuredKong =
+          if (usesAuth)
+            Http("http://localhost:8000/api/auth/login")
+              .method("POST")
+              .asString
+              .code == HttpURLConnection.HTTP_BAD_REQUEST
+          // TODO: this might not be true, needs investigating
+          else true
```
xsanda, line 38: so if we don’t use auth, we still exec, but we don’t wait to check its consequences?

xsanda, line 38: so if we don’t use auth, we still exec, but we don’t wait to check its consequences?

c-harding (commented):
now I look at this I’m not sure it’s your fault, but there seem to be some weirdness around the kong configuration

c-harding (approved):
Looks good to me

PR #296, opened 2020-04-09
[go-gen] Fix interservice comms service names and config.json URL

### LINKED ISSUE #289: Foreign Keys don't work
Minimal example:
```
Foreign: project{
  #language(go);
  #database(postgres);
  #provider(dockerCompose);
}

Foo: service {
  field: int;
  bar: Bar;
}

Bar: service {
  test: int;
}
```

When trying to create a new `Foo`, you first need to create a new `Bar`. This works fine, but the CREATE on the `Foo` checks for the existence of the `Bar`, which comes back as negative, even though the `Bar` does exist, running a manual `GET` to the `/api/bar/{UUID}` endpoint returns the expected result.

willjvsmith:
* Makes the service names in the `comms` field of `ServiceRoot` of type `ServiceName`
  * Makes `ServiceRoot` extend `ServiceName` so that we don't have to do `ServiceRoot.name.name`
* Fixes `config.json` generation of inter-service comms URLS
* Updates `config.json` generation to include service names in kebab case

Fixes #289, I need a break


jaylees14 (approved):

c-harding (approved):
Lgtm

PR #297, opened 2020-04-09
[Validator] Add cycle detection to the validator, using Tarjan’s algo

c-harding:
This uses Tarjan’s algorithm to break the foreign key dependency graph into strongly connected components. If any SCC contains more than one node, there is a cycle, and so an error is raised.


jaylees14 (approved):
Nice job

Lewiky (approved):
Great job, should be good to write about

PR #299, opened 2020-04-10
[go-gen] Fix issue when generating only datetime

### LINKED ISSUE #298: Datetimes are not correctly generated unless accompanied with another date or time attribute
The following Templefile, where a single `datetime` attribute is provided is not correctly generated. For some reason the conversion from `string → time.Time` is omitted in Go. 
```
DateyBoi: project {
  #provider(dockerCompose);
}

User: service {
  dt: datetime;
}
```

However, if you add an additional field such as `time` or `date`, it is correctly generated for both types:

```
DateyBoi: project {
  #provider(dockerCompose);
}

User: service {
  dt: datetime;
  t: time;
}
```

jaylees14:
Fixes #298, also adds integration test to check it works.


Lewiky (approved):
Love it

willjvsmith (approved):
Smh

PR #301, opened 2020-04-10
Comms set

### LINKED ISSUE #300: Duplicate foreign keys are not removed
Duplicate foreign keys cause comm/handler functions to also be duplicated in Go. Imagine it's a `Seq` vs `Set` issue somewhere

MRE:
```
Tinder: project {
  #language(go);
  #database(postgres);
  #provider(dockerCompose);
}

User: service {
  firstName: string;
}

Match: service {
  userOne: User;
  userTwo: User;
}
```

jaylees14:
All I wanted to do today was add `list` support into `temple test`, but I keep hitting problems 😢 

This fixes #300, where we treated foreign keys as a `seq`, rather than a set, causing duplication of code. Have also updated tests to include foreign keys, now we support them.

### DIFF
--- a/src/test/scala/temple/builder/DatabaseBuilderTestData.scala
+++ b/src/test/scala/temple/builder/DatabaseBuilderTestData.scala
@@ -5,7 +5,7 @@ import temple.generate.database.ast.ColumnConstraint.{Check, NonNull, Unique}
 import temple.generate.database.ast.ComparisonOperator.{GreaterEqual, LessEqual}
 import temple.generate.database.ast.Condition.PreparedComparison
 import temple.generate.database.ast.Statement._
-import temple.generate.database.ast._
+import temple.generate.database.ast.{ColumnDef, _}
 
 object DatabaseBuilderTestData {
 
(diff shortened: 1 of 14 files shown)


Lewiky (approved):
LGTM

src/test/scala/temple/builder/DatabaseBuilderTestData.scala
```diff
@@ -5,7 +5,7 @@ import temple.generate.database.ast.ColumnConstraint.{Check, NonNull, Unique}
 import temple.generate.database.ast.ComparisonOperator.{GreaterEqual, LessEqual}
 import temple.generate.database.ast.Condition.PreparedComparison
 import temple.generate.database.ast.Statement._
-import temple.generate.database.ast._
+import temple.generate.database.ast.{ColumnDef, _}
```
c-harding: huh?

c-harding (commented):
looks good, nice and simple

c-harding (approved):

src/test/scala/temple/builder/DatabaseBuilderTestData.scala
jaylees14: I'm not sure either, shall fix

PR #302, opened 2020-04-12
[sql-gen] Use snake case for temple defined attributes

jaylees14:
Fix Postgres gen so it's `created_by` not `createdBy`


Lewiky (approved):

c-harding (approved):
`very_nice`

PR #304, opened 2020-04-13
[project-test] Use temple test as an integration test

jaylees14:
I gift `temple test` as an integration test.

Very simple usage: just check that no exceptions were thrown.
Probably should only use with `docker-compose` for speed, not sure how well `Kubernetes` will work on Travis either.

Also:
- Found a typo in some of the min/max logic, so I fixed that.
- Added a consistent timeout to each integration test


Lewiky (approved):
LGTM

src/it/scala/temple/containers/EndpointTesterSpec.scala
```diff
@@ -0,0 +1,32 @@
...
+import temple.DSL.semantics.Analyzer
+import temple.builder.project.ProjectBuilder
+import temple.detail.LanguageSpecificDetailBuilder
+import temple.test.ProjectTester
+import temple.utils.FileUtils
+import temple.utils.MonadUtils.FromEither
+
+abstract class EndpointTesterSpec extends FlatSpec {
+
+  def testEndpoints(templefile: String): Unit = {
```
c-harding: `buildAndTestEndpoints`? Reading the test, it sounds like it tests something that’s already been set up

src/it/scala/temple/containers/HadolintSpec.scala
```diff
@@ -19,6 +19,10 @@ abstract class HadolintSpec extends DockerShell2HttpService(8080) with DockerTes
   // Validate a given Dockerfile, returning the output of Hadolint
   def validate(dockerfile: String): String = {
     val json = Map("contents" -> dockerfile).asJson.toString()
-    Http(hadolintVerifyUrl).params(Map("dockerfile" -> json)).asString.body
+    Http(hadolintVerifyUrl)
+      .params(Map("dockerfile" -> json))
+      .timeout(connTimeoutMs = 1000, readTimeoutMs = 30000)
+      .asString
+      .body
```
c-harding, line 26: This pattern is used so often, why not add it as a method in DockerShell2HttpService?

src/main/scala/temple/ast/AttributeType.scala
```diff
@@ -42,7 +42,7 @@ object AttributeType {
       else Float.MaxValue
     }
 
-    def minValue: Double = max.getOrElse {
+    def minValue: Double = min.getOrElse {
```
c-harding, line 45: My bad, thank you

c-harding (approved):
nice

src/it/scala/temple/containers/EndpointTesterSpec.scala
jaylees14: Good idea

src/it/scala/temple/containers/HadolintSpec.scala
jaylees14: Shall do in a follow up

PR #305, opened 2020-04-13
[go-gen] Add datastore interface to allow for extension

jaylees14:
This renames the existing DAO interface to `BaseDatastore`, and creates an empty interface that extends this interface called `Datastore`.

This allows users of temple to add their own DAO functions _and_ have them accessible from within hooks.

@Lewiky this is one of the files that shouldn't be regenerated


Lewiky (approved):
COOL

src/e2e/resources/simple-temple-expected/auth/dao/datastore.go
```diff
@@ -0,0 +1,6 @@
+package dao
+
+// Datastore provides the interface adopted by the DAO, allowing for mocking
+type Datastore interface {
+	BaseDatastore
+}
```
c-harding, line 6: Inheritance is dumb

c-harding (approved):
very smart

PR #306, opened 2020-04-13
[sql-gen] Snake case user provided attributes

jaylees14:
As per Postgres standards (https://stackoverflow.com/a/2878408/) (StackOverflow counts as standard, right?), write all field names in `snake_case`


Lewiky (approved):
Cool coolcoolcool

src/main/scala/temple/builder/DatabaseBuilder.scala
```diff
@@ -59,10 +59,15 @@ object DatabaseBuilder {
...
-    val tableName = StringUtils.snakeCase(serviceName)
-    val columns   = attributes.keys.map(Column).toSeq
+    val tableName = snakeCase(serviceName)
+    val columns   = attributes.keys.map(snakeCase(_)).map(Column).toSeq
     val providedColumns =
-      attributes.filter { case (_, attr) => attr != IDAttribute && attr != CreatedByAttribute }.keys.map(Column).toSeq
+      attributes
+        .filter { case (_, attr) => attr != IDAttribute && attr != CreatedByAttribute }
+        .keys
+        .map(snakeCase(_))
```
c-harding: Is there a reason why you’ve included the underscore here, rather than just the bare method name? Or conversely, why the maps were kept separate rather than `Column(snakeCase(_))`?

c-harding (approved):
Beautifully snaky

willjvsmith (approved):

src/main/scala/temple/builder/DatabaseBuilder.scala
jaylees14: `snakeCase` takes two arguments (albeit one with a default value), so you can't just use the bare method name.

`Column(snakeCase(_))` sadly doesn't work without naming the argument, so `(name => Column(snakeCase(name))` if you think that's nicer?

PR #307, opened 2020-04-13
[go-gen] Add setup.go and invoke from main function

jaylees14:
This PR does 2 main things:
- Add `setup.go`, which includes a `setup` function for users to add their own hooks / logic / routes etc.
- Refactor the router so it is no longer defined on the environment, instead it is returned and passed to the user for modification

### Test Plan
- Existing integration tests cover that the code complies and endpoints still produce correct responses

### DIFF
--- a/src/e2e/resources/simple-temple-expected/auth/auth.go
+++ b/src/e2e/resources/simple-temple-expected/auth/auth.go
@@ -48,8 +48,8 @@ type loginAuthResponse struct {
 	AccessToken string
 }
 
-// router generates a router for this service
-func (env *env) router() *mux.Router {
+// defaultRouter generates a router for this service
+func defaultRouter(env *env) *mux.Router {
 	r := mux.NewRouter()
 	r.HandleFunc("/auth/register", env.registerAuthHandler).Methods(http.MethodPost)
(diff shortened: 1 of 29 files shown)


src/e2e/resources/simple-temple-expected/auth/auth.go
```diff
@@ -48,8 +48,8 @@ type loginAuthResponse struct {
 	AccessToken string
 }
 
-// router generates a router for this service
-func (env *env) router() *mux.Router {
+// defaultRouter generates a router for this service
+func defaultRouter(env *env) *mux.Router {
```
c-harding, line 52: When would a router not be a default router?

c-harding (approved):
All the power to Captain Hook!

Lewiky (approved):
So many changes for such little extra logic

src/e2e/resources/simple-temple-expected/auth/auth.go
jaylees14: The router gets given to the user for modification, this just defaults to everything we've generated

willjvsmith (approved):
LGTM 🚀 

The `Call into non-generated entry-point` comment seems a bit funny, since the `env.setup` could theoretically do more than add an entry point?

PR #308, opened 2020-04-13
[Codecov] Include e2e tests in the codecov report

c-harding:
Instead of the unit testing stage pushing the codecov report, instead the e2e stage also calls the unit tests, and pushes the combined codecov report. This means that the e2e stage will fail if the unit tests fail, which I don’t think is too bad of a side effect. They need to be run together somehow in order to generate a single result, and this seems less misleading than the unit testing stage failing because an e2e test is bad.


jaylees14 (approved):
LGTM, look at that codecov 🚀

Lewiky (approved):
LGTM

jaylees14 (approved):

PR #311, opened 2020-04-14
[go-gen] Make metric.go conditionally generated

willjvsmith:
* Makes metric.go only generate if the service should have metrics, including auth service

Simple stuff


jaylees14 (approved):
Nice

c-harding (approved):

PR #312, opened 2020-04-14
[go-gen] Extract DAO input to separate variable

jaylees14:
Extract the DAO input into a separate variable.

This allows a hook call to be nicely squeezed in the middle (à la TempleEight/spec-golang#80)

This is currently only for non-auth services, will do auth next


willjvsmith (approved):
LGTM 🚀

Lewiky (approved):

PR #313, opened 2020-04-14
[go-gen] Extract auth DAO input to separate variable

jaylees14:
Applying the same technique from #312 to the auth service. Extract input into variable so that we can squeeze a hook in the middle

### DIFF
--- a/src/e2e/resources/simple-temple-expected/auth/auth.go
+++ b/src/e2e/resources/simple-temple-expected/auth/auth.go
@@ -129,11 +129,13 @@ func (env *env) registerAuthHandler(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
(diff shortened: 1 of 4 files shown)


willjvsmith (commented):
How is this passing without any scala changes?

jaylees14:
> How is this passing without any scala changes?

You should ask the man that wrote it... 😉 It's just read in from file

https://github.com/TempleEight/temple/blob/4c190195d0f4c95833f05087ce9a97d78da47574/src/main/scala/temple/generate/server/go/auth/GoAuthServiceMainGenerator.scala#L34

willjvsmith (approved):

willjvsmith:
genius

Lewiky (approved):

PR #314, opened 2020-04-14
[Semantics] Make lookupMetadata more typesafe: limit to project metadata

c-harding:
Now `lookupMetadata` can only be called on metadata that is allowed to appear in the project block, so non-inheritable metadata cannot be looked up. The only catch with this is that `structBlock.lookupMetadata` can’t be used for metadata such inherited from the service to the struct. Not sure whether this is desirable, but if it is then a method could be added to `AttributeBlock`: `def lookupMetadata[T <: ServiceMetadata]: Option[T]`

Codecov only down due to reduction in LoC


Lewiky (approved):
LGTM

jaylees14 (approved):

PR #315, opened 2020-04-14
Invoke hooks

jaylees14:
I just ❤️ hooks - this adds "before" hooks to all endpoints.

The only caveat is that if a hook fails - that is to say it throws an error - that error does not return a response.

Once @smithwjv has created the `respondWithError` function, that should slot nicely into the TODO spot I've left. For now it's easier to leave it blank than spend time working out what should go here, since he's going to get that done fairly soon.


willjvsmith (approved):
Good stuff 🚀

Lewiky (approved):
Cool nice one

PR #316, opened 2020-04-14
[go-gen] Add generation of metric server instantiation in main, and refactor main gen

willjvsmith:
* Adds generation of the code in the main functions to spin up a Prometheus metric server
  * For both auth and non-auth services
* Tidies up the main function generation
* Adds Ports to the `util.go` `config` struct

You will notice I've commented out metric imports ready for my next PR where I add the actual metric calls to the handlers.

### DIFF
--- a/src/main/scala/temple/generate/server/go/common/GoCommonMainGenerator.scala
+++ b/src/main/scala/temple/generate/server/go/common/GoCommonMainGenerator.scala
@@ -29,57 +29,102 @@ object GoCommonMainGenerator {
       ),
     )
 
-  private[go] def generateMain(serviceName: ServiceName, port: Int, usesComms: Boolean, isAuth: Boolean): String =
+  private def generateFlagParseBlock(serviceName: ServiceName): String =
+    mkCode.lines(
+      genDeclareAndAssign(
+        genMethodCall(
+          "flag",
+          "String",
+          doubleQuote("config"),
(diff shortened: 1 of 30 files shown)


willjvsmith:
Something fishy is causing the integration tests to fail, I'm not certain, but I don't think it has anything to do with the contents of this PR.

The output of the compiler should be silenced, but for some reason isn't behaving that way.

Lewiky (approved):
Cool cool cool

src/main/scala/temple/generate/server/go/common/GoCommonMainGenerator.scala
```diff
@@ -29,57 +29,102 @@ object GoCommonMainGenerator {
...
+    )
+
+  private def generateMetricsBlock(): String = {
+    val goroutineBody = mkCode.lines(
+      genMethodCall("http", "Handle", doubleQuote("/metrics"), genMethodCall("promhttp", "Handler")),
+      genMethodCall("http", "ListenAndServe", genMethodCall("fmt", "Sprintf", doubleQuote(":%d"), "promPort"), "nil"),
+    )
+    mkCode.lines(
+      "// Prometheus metrics",
+      genDeclareAndAssign(s"config.Ports[${doubleQuote("prometheus")}]", "promPort", "ok"),
```
jaylees14, line 66: `)}]` this is very pleasing

jaylees14 (approved):
Nice

PR #319, opened 2020-04-14
[DSL] Move auth method to be specified in the project block

c-harding:
Instead of `#auth(method)` in the service block, we now have `#authMethod(method)` in the project block, and bare `#auth` in the service block

### DIFF
--- a/src/main/scala/temple/ast/Metadata.scala
+++ b/src/main/scala/temple/ast/Metadata.scala
@@ -92,12 +92,15 @@ object Metadata {
       Omit(names.map(Endpoint.parse(_)).toSet)
   }
 
-  sealed abstract class ServiceAuth private (name: String) extends EnumEntry(name) with ServiceMetadata
+  sealed abstract class AuthMethod private (name: String) extends EnumEntry(name) with ProjectMetadata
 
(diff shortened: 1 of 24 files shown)


jaylees14 (approved):
Nice one

src/main/scala/temple/ast/Metadata.scala
```diff
@@ -92,13 +92,16 @@ object Metadata {
...
 
-  object ServiceAuth extends Enum[ServiceAuth] {
-    override def values: IndexedSeq[ServiceAuth] = findValues
-    case object Email extends ServiceAuth("email")
+  object AuthMethod extends Enum[AuthMethod] {
+    override def values: IndexedSeq[AuthMethod] = findValues
+    case object Email extends AuthMethod("email")
   }
 
+  type ServiceAuth = ServiceAuth.type
```
Lewiky, line 102: Why is this line needed? And it's defined before `ServiceAuth` is?

Lewiky (approved):
Look good

src/main/scala/temple/ast/Metadata.scala
c-harding: So `lookupMetadata` etc work nicely. Order doesn’t really matter

PR #320, opened 2020-04-14
[it-test] Add examples folder and use in integration test

jaylees14:
Move the `tinder.temple` over (`amazon.temple` didn't generate correctly, will move over when I've worked out why..)

Invoke them from the integration test automatically, so that any new examples are automatically generated and validated.


willjvsmith (approved):
Good stuff

Lewiky (approved):
LGTM

src/it/scala/temple/EndpointIntegrationTest.scala
```diff
@@ -11,4 +13,11 @@ class EndpointIntegrationTest extends EndpointTesterSpec with Matchers {
     val simpleTemple = FileUtils.readFile("src/test/scala/temple/testfiles/simple-dc.temple")
     noException should be thrownBy buildAndTestEndpoints(simpleTemple)
   }
+
+  it should "correctly generate the example Templefiles" in {
+    Files.list(Paths.get("examples")).map(path => path.toString).forEach { (path: String) =>
```
c-harding: Worth filtering for `*.temple` so we can add a readme to that folder without it being imported?

examples/tinder.temple
```diff
@@ -0,0 +1,26 @@
...
+  #database(postgres);
+  #provider(dockerCompose);
+}
+
+User: service {
+  // Implicit `id` field that matches Auth service `id` field
+  firstName: string;
+  lastName: string;
+  dob: datetime;
+  #auth(email); // Creates Auth service, which stores an `id`, `email` and `password` field
```
c-harding: #319 will break this, race you 😉

c-harding (approved):

PR #321, opened 2020-04-14
[go-gen] Add generation of timing and request success metrics

willjvsmith:
* Adds generation of metrics for timing DAO calls
* Adds generation of metrics for counting the success of a request

This is for both auth and non-auth services. I moved the metric generation functions to `common`, then realised that the handlers are read from file in `auth` atm 🤦 

This was really goddamn tedious, and the refactor with `respondWithError` is going to be even worse.

### DIFF
--- a/src/main/scala/temple/generate/server/go/service/main/GoServiceMainListHandlerGenerator.scala
+++ b/src/main/scala/temple/generate/server/go/service/main/GoServiceMainListHandlerGenerator.scala
@@ -4,7 +4,7 @@ import temple.generate.CRUD.List
 import temple.generate.server.ServiceRoot
 import temple.generate.server.go.GoHTTPStatus.StatusInternalServerError
 import temple.generate.server.go.common.GoCommonGenerator._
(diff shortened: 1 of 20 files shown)


src/main/scala/temple/generate/server/go/service/main/GoServiceMainListHandlerGenerator.scala
```diff
@@ -4,19 +4,22 @@ import temple.generate.CRUD.List
...
+import temple.generate.server.go.common.GoCommonMainGenerator._
+import temple.generate.server.go.service.main.GoServiceMainHandlersGenerator._
 import temple.generate.utils.CodeTerm.{CodeWrap, mkCode}
 
 import scala.Option.when
 import scala.collection.immutable.ListMap
 
 object GoServiceMainListHandlerGenerator {
 
+  // TODO: This needs a refactor
```
jaylees14, line 16: 👍 thought the same

jaylees14 (approved):
Nice stuff 👍

Lewiky (approved):
Cool one

PR #324, opened 2020-04-15
[Go-gen] Move ServiceRoot to a trait in preparation for StructRoot

c-harding:
This rearranges the ServiceRoot structure to prepare for the introduction of structs, and also fixes the notion of equality in ServiceNames that was breaking some tests. As such, it fixes those tests.

Some tests were made less non-deterministic by swapping ListMaps (which were constructed from unordered sets) with SortedMaps, which derive their order from CRUD’s internal ordering.

## Old inheritance structure

    class ServiceName
    ┃   overriding equality
    ┃   declaring name
    ┃   defining additional name transformers
    ┗ case class ServiceRoot
          defining name, as well as service-specific stuff
          not generating equality, see below †

## New structure:

    trait ServiceName
    ┃   declaring name                                                          
    ┃   defining additional name transformers
    ┣ private case class ServiceName.BareName
    ┃     constructed with ServiceName(), acts like a ServiceName case class as before
    ┃     generates equality
    ┗ trait AttributesRoot
      ┃   still declaring name
      ┃   declaring service/struct shared stuff
      ┡ case class ServiceRoot
      ┆     defining name and service/struct shared stuff
      ┆     defining service-specific stuff
      └ case class StructRoot (coming soon)
            defining name and service/struct shared stuff
            defining struct-specific stuff

† Case classes normally generate new equality functions, but annoyingly they do not if they extend a class that already overrides equality. This goes against an assumption previously made that ServiceName would generate a new equality method


src/main/scala/temple/generate/server/AttributesRoot.scala
```diff
@@ -0,0 +1,60 @@
...
+    createdByAttribute: Option[CreatedByAttribute],
+    attributes: ListMap[String, AbstractAttribute],
+    datastore: Database,
+    readable: Readable,
+    writable: Writable,
+    projectUsesAuth: Boolean,
+    hasAuthBlock: Boolean,
+    metrics: Option[Metrics],
+  ) extends ServiceName
+      with AttributesRoot {}
```
jaylees14: Do you need the `{}`

jaylees14 (approved):
LGTM

src/main/scala/temple/generate/server/go/service/GoServiceGenerator.scala
```diff
@@ -49,20 +44,29 @@ object GoServiceGenerator extends ServiceGenerator {
       File(root.kebabName, s"${root.kebabName}.go") -> mkCode.doubleLines(
         GoCommonGenerator.generatePackage("main"),
         GoServiceMainGenerator
-          .generateImports(root, usesBase64, usesTime, usesComms, usesMetrics, clientAttributes, operations),
+          .generateImports(root, usesBase64, usesTime, usesComms, usesMetrics, root.requestAttributes, root.operations),
```
willjvsmith: perhaps could do with refactoring now that requestAttributes and operations are defined on the root, so only the root and the bools are passed

willjvsmith (approved):
LGTM 🚀 

good stuff

PR #325, opened 2020-04-15
[go-gen] Add respondWithError function gen, and use in auth service

willjvsmith:
* Adds `respondWithError` function generation to both auth and non-auth services
  * Handles metrics/no metrics
* Replaces `errMsg...http.Error` statements with `respondWithError` in auth service
  * Including hook TODOs
* Refactors auth service import generation to make it easier to add and remove imports, if a bit more verbose

Will replace in non-auth services in following PR


c-harding (approved):
very nice

jaylees14 (approved):
Nice one 👍

PR #326, opened 2020-04-15
[openapi-gen] Add identity endpoint

jaylees14:
The first of N PR's where I add a new `GET` endpoint.
Here I have:
- Added the `identity` field to the CRUD enum
- Fixed any direct errors that were as a result of this (namely the validator)
- Added the endpoint to the OpenAPI spec, so we can discuss if this follows what we'd expect to be generated.

Whilst updating the OpenAPI I noticed that the `BodyLiteral` being used should only be for requests and not for response bodies (since it doesn't contain headers, and requires content to be present), so I have created the new objects as per the OpenAPI spec. 

Additionally I added an integration test which validates what we are doing is valid OpenAPI.


src/it/scala/temple/generate/target/openapi/OpenAPIGeneratorIntegrationTest.scala
```diff
@@ -1,7 +1,14 @@
 package temple.generate.target.openapi
 
 import org.scalatest.Matchers
+import temple.ast.AbstractAttribute.Attribute
+import temple.ast.Annotation
+import temple.ast.AttributeType.{BlobType, BoolType, DateTimeType, DateType, FloatType, ForeignKey, IntType, StringType, TimeType}
```
willjvsmith:
```suggestion
import temple.ast.AttributeType._
```

src/test/scala/temple/generate/target/openapi/OpenAPIGeneratorTest.scala
```diff
@@ -178,6 +178,20 @@ class OpenAPIGeneratorTest extends FlatSpec with Matchers {
...
+        |      responses:
+        |        '302':
+        |          description: The single match is accessible from the provided Location
+        |          headers:
+        |            Location:
+        |              description: The location where the single match can be found
+        |              schema:
+        |                type: string
+        |        '404':
+        |          $ref: '#/components/responses/Error404'
```
willjvsmith, line 194: Shouldn't this endpoint only generate for services with an auth block? i.e. `User` not `Match`

Apologies if this test differs from the spec-golang model

jaylees14: Yeah the "match" name is somewhat incidental, the test just generates every possible endpoint for a service with every field type.

jaylees14: The logic to decide when to generate this endpoint isn't yet implemented - this is evident from the lack of changes to any `E2E` test 😄

Lewiky (approved):
LGTM

willjvsmith (approved):
LGTM 🚀

PR #328, opened 2020-04-15
[go-gen] Add respondWithError to non-auth handlers

willjvsmith:
* Renames `generateRespondWithError` function to `generateRespondWithErrorFunc` to avoid confusion with handler `generateRespondWithError` function
* Replaces `errMsg...http.Error` blocks with `respondWithError` in non-auth services
  * Adds it to the hook TODOs too
* Updates an ungodly number of test lines

Apologies for the size, there is no easy way to make it smaller as all the handlers share generator functions.

To give some intuition before you read the code, I have essentially replaced the `generateHTTPError` functions with `generateRespondWithError` functions.

Since `respondWithError` must support metrics, I pass around a `metricSuffix: Option[String]` which is `None` if the project has no metrics, and `"Create"` or `"List"` for example if there are metrics. This is created at each CRUD handler generator. Why not just pass around a CRUD op you ask? I wrote this keeping structs and other metrics in mind, e.g. `"DeletePicture"`.

You may also wonder why I replaced the `GoHTTPStatus` argument in `generateRespondWithError` with a string. This is because in the hooks, a custom error can be returned, so `"err.statusCode"` is given, which is not a `GoHTTPStatus`. I've created a helper function to bridge the gap: `genHTTPEnum`.


src/main/scala/temple/generate/server/go/service/main/GoServiceMainHandlersGenerator.scala
```diff
@@ -48,46 +48,30 @@ object GoServiceMainHandlersGenerator {
...
     s"func (env *env) ${operation.toString.toLowerCase}${root.name}Handler(w http.ResponseWriter, r *http.Request)"
 
-  /** Generate a errMsg declaration and http.Error call */
-  private[main] def generateHTTPError(statusCodeEnum: GoHTTPStatus, errMsg: String, errMsgArgs: String*): String = {
-    val createErrorJSONargs =
+  /** Generate a respond with error call */
+  private[main] def generateRespondWithError(
+    statusCode: String,
+    metricSuffix: Option[String],
+  )(errMsg: String, errMsgArgs: String*): String = {
```
jaylees14, line 55: I'm struggling to see why splitting this into two sets of arguments is useful? Why aren't all the arguments taken together?

jaylees14 (approved):
LGTM

src/main/scala/temple/generate/server/go/service/main/GoServiceMainHandlersGenerator.scala
willjvsmith: You're right, I originally didn't have MetricSuffix as an optional, just as a string, and was overloading by omitting it.

c-harding (approved):
looks good, very repetitive but I don’t think that can easily be avoided

PR #338, opened 2020-04-16
[project-test] Use identify endpoint to avoid access token logic

jaylees14:
So.. suppose we have a match service like this:
```
Match: service {
  userA: User;
  userB: User;
}
```

If we're trying to mock a request body for this, we previously assumed that we need to create two users. To do this, we would create 2 new access tokens, and make 2 new users (since an access token can only be associated with one user).

However, you can only create a match here because the `User` service was `#readable(by: all)` and therefore it was possible to know about other users.

This doesn't work in the `amazon.temple` example, whereby `Customers` are only accessible by themselves (so `#readable(by: this)`). This means the access token that creates an `Order` (associated with a customer) may only be created by the same token that created the `Customer`.

So, I added a new endpoint that allows you to look up a user ID from an access token - which then allows you to correctly populate request bodies with users associated with tokens.

This removes all the gross logic to do with throwaway access tokens: we now just create 1 token per endpoint, and use that for all requests. 

This does come with the caveat that in the `tinder.temple` example, UserA == UserB, but there's no reason that this shouldn't be allowed.

### DIFF
--- a/src/main/scala/temple/test/internal/CRUDServiceTest.scala
+++ b/src/main/scala/temple/test/internal/CRUDServiceTest.scala
@@ -14,8 +14,11 @@ class CRUDServiceTest(
   usesAuth: Boolean,
 ) extends ServiceTest(name) {
 
-  private def testCreateEndpoint(accessToken: String): Unit =
+  private def newToken(): String = if (usesAuth) ServiceTestUtils.getAuthTokenWithEmail(name, baseURL) else ""
+
+  private def testCreateEndpoint(): Unit =
     testEndpoint("create") { test =>
+      val accessToken = newToken()
       val requestBody =
         ServiceTestUtils.constructRequestBody(test, service.attributes, allServices, baseURL, accessToken)
       val createJSON =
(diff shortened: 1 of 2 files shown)


src/main/scala/temple/test/internal/ServiceTestUtils.scala
```diff
@@ -151,34 +151,54 @@ object ServiceTestUtils {
...
+  ): String = {
     val service = allServices.getOrElse(serviceName, test.fail(s"service $serviceName does not exist"))
-    // If this service is an auth service, the same access token cannot be used twice, so make a new one to be safe...
-    // This is so that services that reference 2+ auth'd services can be successfully tested
-    // TODO: refactor for central auth method
-    val newAccessToken = service.lookupLocalMetadata[Metadata.ServiceAuth].fold(accessToken) { _ =>
-      ServiceTestUtils.getAuthTokenWithEmail(serviceName, baseURL)
+
+    // If this service is an auth service, an access token can only create one entity
+    val createJSON = service.lookupLocalMetadata[Metadata.ServiceAuth] match {
```
c-harding: `hasMetadata` works for boolean metadata, may be slightly clearer

```scala
val createJSON = if (service hasMetadata Metadata.ServiceAuth) {
} else {
}
```

src/main/scala/temple/test/internal/ServiceTestUtils.scala
```diff
@@ -151,34 +151,54 @@ object ServiceTestUtils {
...
+        // First check if that single entity already exists
+        val response = Http(s"http://$baseURL/api/${StringUtils.kebabCase(serviceName)}")
+          .method("GET")
+          .header("Authorization", s"Bearer $accessToken")
+          .asString
+        response.code match {
+          case HttpURLConnection.HTTP_MOVED_TEMP =>
+            // Already exists, so perform a GET on it now the ID is known
+            val url = response.header("Location").getOrElse(test.fail("302 response without Location header"))
+            getRequest(test, s"http://$url", accessToken)
```
c-harding: This feels like a broken url format if you have to stick `http://` onto it

src/main/scala/temple/test/internal/CRUDServiceTest.scala
```diff
@@ -24,58 +27,56 @@ class CRUDServiceTest(
...
+  private def testReadEndpoint(): Unit =
     testEndpoint("read") { test =>
+      val accessToken    = newToken()
       val createResponse = ServiceTestUtils.create(test, name, allServices, baseURL, accessToken)
       val getJSON = ServiceTestUtils
         .getRequest(
           test,
-          s"http://$baseURL/api/${StringUtils.kebabCase(name)}/${createResponse.id}",
-          createResponse.accessToken,
+          s"http://$baseURL/api/${StringUtils.kebabCase(name)}/${createResponse}",
```
c-harding, line 37: What was the reason for stripping the `http://` from the `baseURL` in performSetup?

c-harding (approved):
Seems very sensible

Lewiky (approved):
LGTM

src/main/scala/temple/test/internal/ServiceTestUtils.scala
jaylees14: Will update the location header in a separate PR.

src/main/scala/temple/test/internal/CRUDServiceTest.scala
jaylees14: That's for providing the host when configuring Kong. https://docs.konghq.com/2.0.x/admin-api/#add-route

willjvsmith (commented):
LGTM 🚀

willjvsmith (approved):

PR #339, opened 2020-04-16
[Testing] Move test snippets to autogenerated map

c-harding:
Not sure why these were stored as snippets, but this should make them more consistent


jaylees14 (approved):
Looks good. For some context, they were snippets because they weren't actually valid Go code until very recently.

Lewiky (approved):
Cool

PR #341, opened 2020-04-16
[project-test] Add explict test for identify endpoint

jaylees14:
Add an explicit test for the identify endpoint

@xsanda I'll fix the `http://` prefix next.


Lewiky (approved):
Cool cool cool

c-harding (approved):
Nice, is it worth testing 404 before creating it?

jaylees14:
@xsanda That's a nice idea. To do that nicely requires a little refactor, so I'll do it in a follow up 😄

PR #342, opened 2020-04-16
[examples] Add amazon example

jaylees14:
Move amazon example over from `example-templefiles`, now it works 😉


Lewiky (approved):

examples/amazon.temple
```diff
@@ -0,0 +1,64 @@
...
+
+Product: service {
+  category: ProductCategory;
+  seller: Seller;
+  name: string;
+  description: string;
+  price: float;
+
+  Picture: struct {
+    product: Product;
```
willjvsmith: Can be omitted because of implicit `parentID`

jaylees14: Yes indeed, nice catch

examples/amazon.temple
```diff
@@ -0,0 +1,63 @@
...
+  customer: Customer;
+  orderDate: datetime;
+
+  // OrderItem is a join table between Orders and Products (many to many)
+  OrderItem: struct {
+    product: Product;
+    #enumerable;
+  }
+
+  #omit([delete]);
```
c-harding: parens are optional with arrays, may look nicer without? Up to you

c-harding (approved):
lgtm

PR #343, opened 2020-04-16
[Regen] Implement Regeneration

Lewiky:
Implements regeneration of code if `temple generate` is called in a directory that already contains some files that were previous generated by temple. Confirms with the user before actually deleting anything. 

Implemented by generating everything as normal, but before writing anything to a file, checking if any of the generated files already exist. If they do and the user wants to overwrite, it simply removes any business logic files (`setup.go` and `datastore.go`) from the list of generated files to be outputted, the rest get overwritten. 

## Test Plan

Unit tests are written, but can test for yourself by trying to generate a templefile in a directory it's already been generated in before, make a change to a `setup.go` and it should still be there after generating again.


Lewiky:
As mentioned in Slack, there's currently an awful hack in the `filesUnderDirectory` function - needs replacing

src/test/scala/temple/RegenerationFilterTest.scala
```diff
@@ -0,0 +1,68 @@
...
+import temple.detail.PoliceSergeantNicholasAngel
+import temple.generate.FileSystem
+
+import scala.reflect.io.Directory
+
+class RegenerationFilterTest extends FlatSpec with Matchers with BeforeAndAfter {
+
+  behavior of "RegenerationFilterTest"
+
+  private val testDirectory = "/tmp/test"
```
jaylees14: I'd be tempted to use `Files.createTempDirectory(...)`?

src/test/scala/temple/RegenerationFilterTest.scala
```diff
@@ -0,0 +1,68 @@
...
+
+  after {
+    val testDirectory = new File(this.testDirectory)
+    if (testDirectory.exists()) new Directory(testDirectory).deleteRecursively()
+  }
+
+  private def setupFileSystemWhole(): Unit = {
```
jaylees14, line 26: Can this not go in a `before` block?

jaylees14 (approved):
LGTM, will have a look at the hack

src/test/scala/temple/RegenerationFilterTest.scala
Lewiky: Didn't know that was a thing!

src/test/scala/temple/RegenerationFilterTest.scala
Lewiky: It's not the same for every test, some you want files to exist and some you don't

jaylees14:
Also just a thought, what happens if they change the go module name between generations? I think if they've changed `setup.go` and added imports they'll break?

src/main/scala/temple/RegenerationFilter.scala
```diff
@@ -0,0 +1,94 @@
...
+    if (isRegen(outputDirectory, existingFiles, project)) {
+      val omittedFiles = filesToOmit(existingFiles)
+      // Get the user to confirm whether we're regenerating
+      if (shouldRegen(omittedFiles, questionAsker)) {
+        Project(project.files.filterNot {
+          case file -> _ => omittedFiles.contains(File(file.folder, file.filename))
+        })
+      } else {
+        // If the user doesn't want to overwrite files, don't output anything
+        Project(Map())
```
jaylees14: Hmm, it still outputs: "Generated project in ..." even if I say no

Lewiky:
> Also just a thought, what happens if they change the go module name between generations? I think if they've changed `setup.go` and added imports they'll break?

@jaylees14 yeah they will break, but there's not really a way around that unless we delve into the file itself, which would be a lot more work

src/main/scala/temple/RegenerationFilter.scala
```diff
@@ -0,0 +1,94 @@
...
+        .asScala
+        .filterNot(path => path.toFile.isDirectory)
+        .map(path => FileSystems.getDefault.getPath(directory).resolve(path))
+        .map(path =>
+          // hack - if the file is in $directory then getParent wont contain "directory/", but it will when the
+          // file is in some subdirectory of $directory, allow for both cases
+          File(path.getParent.toString.replace(directory + "/", "").replace(directory, ""), path.getFileName.toString),
+        )
+        .toSeq
+    } else Seq()
```
jaylees14: How about 

```scala
private def filesUnderDirectory(directory: String): Seq[File] = {
    val directoryPath = Paths.get(directory)
    if (Files.exists(directoryPath)) {
      Files
        .walk(directoryPath, 4)
        .iterator()
        .asScala
        .filterNot(path => path.toFile.isDirectory)
        .map(path => directoryPath.relativize(path))
        .map(path => File(Option(path.getParent).fold("")(_.toString), path.getFileName.toString))
        .toSeq
    } else Seq()
  }
```

src/main/scala/temple/RegenerationFilter.scala
```diff
@@ -0,0 +1,94 @@
...
+  private def isRegen(outputDir: String, existingFiles: Seq[File], project: Project): Boolean = {
+    val generatedFiles = project.files.keys.map(file => File(file.folder, file.filename)).toSet
+    existingFiles.toSet.intersect(generatedFiles).nonEmpty
+  }
+
+  // Confirm with the user that they want to overwrite existing files
+  private def shouldRegen(omittedFiles: Seq[File], questionAsker: QuestionAsker): Boolean = {
+    val omittedFileString = mkCode.lines(omittedFiles.map(file => s"➤ ${file.toString}"))
+    var answer            = ""
+    while (!Seq("y", "n").contains(answer.toLowerCase)) {
```
c-harding: `"yes"`

src/main/scala/temple/RegenerationFilter.scala
```diff
@@ -0,0 +1,94 @@
...
+    questionAsker: QuestionAsker,
+  ): Project = {
+    val existingFiles = filesUnderDirectory(outputDirectory)
+    // If we are regenerating, i.e some files already exist here
+    if (isRegen(outputDirectory, existingFiles, project)) {
+      val omittedFiles = filesToOmit(existingFiles)
+      // Get the user to confirm whether we're regenerating
+      if (shouldRegen(omittedFiles, questionAsker)) {
+        Project(project.files.filterNot {
+          case file -> _ => omittedFiles.contains(File(file.folder, file.filename))
```
c-harding: is this just `omittedFiles.contains(file)`?

src/main/scala/temple/RegenerationFilter.scala
c-harding: I’d be tempted to say that this should return an Option

src/test/scala/temple/RegenerationFilterTest.scala
```diff
@@ -0,0 +1,68 @@
...
+    new File(testDirectory).mkdirs
+    val testFile = new File(testDirectory + "/test-file.go")
+    new PrintWriter(testFile).write("foo")
+    val setupGo = new File(testDirectory + "/setup.go")
+    new PrintWriter(setupGo).write("bar")
```
c-harding: Can we use `FileUtils.writeToFile(testDirectory + "/setup.go", "bar")`? It’s what it was created for, and closes the file for you

src/test/scala/temple/RegenerationFilterTest.scala
```diff
@@ -0,0 +1,68 @@
...
+  }
+
+  it should "not filter when no files exist" in {
+    setupFileSystemEmpty()
+    val output = RegenerationFilter.filter(
+      testDirectory,
+      RegenerationFilterTestData.simpleTestProjectWithConflict,
+      PoliceSergeantNicholasAngel,
+    )
+    output shouldBe RegenerationFilterTestData.simpleTestProjectWithConflict
```
c-harding: `projectFilesShouldMatch()`, to get the diff displaying properly on error? And in a few other places too.

src/test/scala/temple/RegenerationFilterTest.scala
```diff
@@ -0,0 +1,68 @@
...
+  }
+
+  it should "do nothing when the user says no" in {
+    setupFileSystemWhole()
+    val output = RegenerationFilter.filter(
+      testDirectory,
+      RegenerationFilterTestData.simpleTestProjectWithConflict,
+      (_: String) => "n", // SAM implementation of a QuestionAsker that always says no
```
c-harding:
```suggestion
      questionAsker = _ => "n", // Reply no when asked to overwrite
```

src/test/scala/temple/RegenerationFilterTestData.scala
```diff
@@ -0,0 +1,14 @@
...
+
+import temple.builder.project.Project
+import temple.generate.FileSystem.File
+
+object RegenerationFilterTestData {
+
+  val simpleTestProjectWithConflict: Project = Project(
+    Map(
+      File("", "test-file.go") -> "test-contents",
+      File("", "setup.go")     -> "setup.go contents",
```
c-harding, line 11: Where does this string come from?

Lewiky: Which string?

src/main/scala/temple/Application.scala
```diff
@@ -24,9 +24,13 @@ object Application {
     val project            = ProjectBuilder.build(analyzedTemplefile, detail)
     val filteredProject    = RegenerationFilter.filter(outputDirectory, project, questionAsker)
 
-    FileUtils.outputProject(outputDirectory, filteredProject)
+    filteredProject match {
+      case Some(project) =>
+        FileUtils.outputProject(outputDirectory, project)
+        println(s"Generated project in $outputDirectory")
+      case None => println("Nothing to generate")
+    }
```
c-harding: Maybe I’m too much of a Swift dev /s, but I feel like the indentation isn’t the nicest way of handling this
```scala
val filteredProject = RegenerationFilter.filter(outputDirectory, project, questionAsker) getOrElse {
  println("Nothing to generate")
  return
}
```

src/main/scala/temple/RegenerationFilter.scala
```diff
@@ -28,21 +27,19 @@ object RegenerationFilter {
...
-        .map(path =>
-          // hack - if the file is in $directory then getParent wont contain "directory/", but it will when the
-          // file is in some subdirectory of $directory, allow for both cases
-          File(path.getParent.toString.replace(directory + "/", "").replace(directory, ""), path.getFileName.toString),
-        )
+        .map(path => directoryPath.relativize(path))
+        .map(path => File(Option(path.getParent).fold("")(_.toString), path.getFileName.toString))
         .toSeq
     } else Seq()
+  }
```
c-harding: How’s this different from `temple.TestUtils#buildFileMap`? Worth moving `buildFileMap` to FileUtils, and just using that?

src/test/scala/temple/RegenerationFilterTestData.scala
c-harding: `"setup.go contents"`

src/main/scala/temple/RegenerationFilter.scala
```diff
@@ -0,0 +1,92 @@
...
+        .filterNot(path => path.toFile.isDirectory)
+        .map(path => directoryPath.relativize(path))
+        .map(path => File(Option(path.getParent).fold("")(_.toString), path.getFileName.toString))
+        .toSeq
+    } else Seq()
+  }
+
+  // Check if any of the files we're generating already exist, i.e if we're regenerating something.
+  private def isRegen(outputDir: String, existingFiles: Seq[File], project: Project): Boolean = {
+    val generatedFiles = project.files.keys.map(file => File(file.folder, file.filename)).toSet
```
c-harding: What does the map achieve?

src/main/scala/temple/RegenerationFilter.scala
```diff
@@ -0,0 +1,92 @@
...
+      val omittedFiles = filesToOmit(existingFiles)
+      // Get the user to confirm whether we're regenerating
+      if (shouldRegen(omittedFiles, questionAsker)) {
+        Some(Project(project.files.filterNot {
+          case file -> _ => omittedFiles.contains(file)
+        }))
+      } else {
+        // If the user doesn't want to overwrite files, don't output anything
+        None
+      }
```
c-harding:
```suggestion
      when(shouldRegen(omittedFiles, questionAsker)) {
        Project(project.files.filterNot {
          case file -> _ => omittedFiles.contains(file)
        })
      }
```

src/test/scala/temple/RegenerationFilterTest.scala
```diff
@@ -1,68 +1,80 @@
...
 
 import scala.reflect.io.Directory
 
-class RegenerationFilterTest extends FlatSpec with Matchers with BeforeAndAfter {
+class RegenerationFilterTest extends FlatSpec with FileMatchers with BeforeAndAfter {
 
   behavior of "RegenerationFilterTest"
 
-  private val testDirectory = "/tmp/test"
+  private var testDirectory: Path = Files.createTempDirectory("test")
```
c-harding:
```suggestion
  private var testDirectory: Path = _
```

This seems like a pointless duplication of `before`

c-harding (approved):
looks good!

src/test/scala/temple/RegenerationFilterTestData.scala
Lewiky: It's just a test value

PR #344, opened 2020-04-16
[Validator] Validate that a struct cannot be more readable than its service

c-harding:
As discussed on Slack, a struct cannot be readable by everyone if its service can only be read by its creator, logistically because the service entry ID is in the URL, and meaningfully because a struct is effectively an attribute of the service, and so should act as such


jaylees14 (approved):
Nice and easy

Lewiky (approved):
LGTM

PR #345, opened 2020-04-16
[grafana] Fix ordering of grafana

jaylees14:
The order of the dashboards in Grafana didn't match our LCRUDI (what an acronym), because of a sorting issue.

Moving the `.toSeq` seems to have fixed.


c-harding (approved):

Lewiky (approved):
Cool cool cool

PR #348, opened 2020-04-17
[go-gen] Fix implementation of blob types

jaylees14:
Fix the implementation of `blob` types in Go, which involves additional conversion from base64 → `[]byte`.

Unfortunately we don't currently check the bounds of the array in the `Go`, only in `Postgres`, so I'll do that in a follow up.

Adds both a new Go integration test & Endpoint test for bounded & unbounded blob types.


c-harding (approved):
Very nice, cheers Jay

willjvsmith (approved):
LGTM 🚀

PR #349, opened 2020-04-17
[go-gen] Use %s in error format string

jaylees14:
Apparently strings are injected with `%s` not `%d`, who knew?!


willjvsmith (approved):
> Apparently strings are injected with %s not %d, who knew?!

Big if true

Lewiky (approved):
LOVE

PR #351, opened 2020-04-17
[metrics] Fix prom address and auth metric names

jaylees14:
Two small fixes:
- The auth service uses `register` and `login` instead of `create` and `read`, so update those
- We named the prometheus service `prometheus`, not `prom`, so update the URL


willjvsmith (approved):
LGTM 🚀

Lewiky (approved):
Nice

PR #353, opened 2020-04-17
[go-gen] Include leading http in response header

jaylees14:
As per https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Location, the location header provided should either be relative to the request URL, or a full absolute URL.

When an absolute URL is used (as per https://tools.ietf.org/html/rfc7231#section-7.1.2), it should include the protocol - so I've updated that here.


Lewiky (approved):
LGTM

c-harding (approved):
very conformant

PR #354, opened 2020-04-17
[test] Allow --testOnly flag to temple test

jaylees14:
You can now provide a `--testOnly` flag to `temple test` and it will only query the endpoints, without spinning up any infra. 

This should hopefully be useful when testing metrics / debugging a database.

Also if anyone has a better name please suggest 😄


c-harding (approved):
looks good!

Lewiky (approved):
Beautiful

PR #355, opened 2020-04-17
[test] Validate kong has been correctly setup when no auth block is present

jaylees14:
Turns out, the TODO needed to be TODONE.

Services without auth may also not correctly configure Kong the first time round, so now I pick a random service and test if we can access it. When we can, Kong is correctly setup.

### DIFF
--- a/src/main/scala/temple/test/ProjectTester.scala
+++ b/src/main/scala/temple/test/ProjectTester.scala
@@ -4,9 +4,11 @@ import java.io.IOException
 import java.net.HttpURLConnection
 
 import scalaj.http.Http
+import temple.ast.AbstractServiceBlock.ServiceBlock
 import temple.ast.Metadata.{AuthMethod, Provider}
 import temple.ast.{Metadata, Templefile}
 import temple.test.internal.{AuthServiceTest, CRUDServiceTest, ProjectConfig}
+import temple.utils.StringUtils.kebabCase
 
 import scala.sys.process._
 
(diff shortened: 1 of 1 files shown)


src/main/scala/temple/test/ProjectTester.scala
```diff
@@ -20,22 +22,29 @@ object ProjectTester {
...
   /** Configure Kong using the generated script, waiting until this is successful */
-  private def configureKong(usesAuth: Boolean, config: ProjectConfig, generatedPath: String): Unit = {
+  private def configureKong(
+    usesAuth: Boolean,
+    services: Map[String, ServiceBlock],
+    config: ProjectConfig,
+    generatedPath: String,
+  ): Unit = {
+    // Get a random service to validate kong is configured correctly
+    val (serviceName, _) = Option(services.head).getOrElse {
```
c-harding: that’s literally what `.headOption` is there for

c-harding (approved):
very clever

src/main/scala/temple/test/ProjectTester.scala
jaylees14: Oh, I must have missed that 😆

src/main/scala/temple/test/ProjectTester.scala
```diff
@@ -20,22 +22,29 @@ object ProjectTester {
     s"sh -c '$command'".!!
 
   /** Configure Kong using the generated script, waiting until this is successful */
-  private def configureKong(usesAuth: Boolean, config: ProjectConfig, generatedPath: String): Unit = {
+  private def configureKong(
```
c-harding, line 25: Does this function add bad entries to the metrics?

willjvsmith (approved):

src/main/scala/temple/test/ProjectTester.scala
jaylees14: Yeah it will do, but it would have also done that before. I'll try and see if `kong` exposes a nicer method to see if things are set up, but this should work for now 😄

PR #352, opened 2020-04-17
[examples] Add 5-a-side templefile

willjvsmith:
* Adds a 5-a-side templefile

This does highlight a few shortcomings, namely auth roles

PR #356, opened 2020-04-17
[go-gen] Add auth arg to hooks

willjvsmith:
* Adds an auth argument to the hooks

This significantly increase the power of the hooks, as it means the hook knows the authID of the request, which can then be checked against other services with #auth blocks, essentially allowing access roles. Since the auth struct is passed, if in the future we add support for roles directly to the auth service, simply updating the util auth struct will make them available to the hooks.

The only access control this does not have the power to solve is creating an instance of a role itself. See the following example from #352:

```
// TODO: There should be a way to restrict who can create a Staff instance, i.e. auth service needs roles
 Staff: service {
   // implicit 'id' field that matches auth service 'id' field
   firstName: string;
   lastName: string;
   email: string;
   #auth;
   // implicit #readable(by: this) since project has auth
   // implicit #writable(by: this) since project has auth
 }

 Player: service {
   // implicit 'id' field that matches auth service 'id' field
   firstName: string;
   lastName: string;
   email: string;
   phoneNumber: int;
   #auth;
   // implicit #readable(by: this) since project has auth
   // implicit #writable(by: this) since project has auth
 }

 /*
  * Strongly enforce only creatable by a player in the beforeCreate hook (loosely enforced by player being readable by
  * this), and only readable by the player or staff in the beforeRead hook
  */
 EmergencyContact: service {
   // implicit 'id' field
   // implicit 'createdBy' field since service does not have auth
   player: Player;
   firstname: string;
   lastName: string;
   relation: string;
   email: string;
   phoneNumber: int;
   #readable(by: all);
   // implicit #writable(by: this) since project has auth
 }
```

Here this PR solves the problem of `EmergencyContact` of only being readable by the `player` that created it and staff. However it does not solve the problem of anyone being to create a new staff instance.

### DIFF
--- a/src/main/scala/temple/generate/server/go/service/GoServiceHookGenerator.scala
+++ b/src/main/scala/temple/generate/server/go/service/GoServiceHookGenerator.scala
@@ -9,6 +9,8 @@ import temple.generate.server.go.common.GoCommonHookGenerator
 import temple.generate.utils.CodeTerm.{CodeWrap, mkCode}
 import temple.utils.StringUtils.doubleQuote
 
+import scala.Option.when
+
 object GoServiceHookGenerator {
 
   private[service] def generateImports(root: ServiceRoot): String = {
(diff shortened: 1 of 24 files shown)


src/main/scala/temple/generate/server/go/service/GoServiceHookGenerator.scala
```diff
@@ -6,35 +6,53 @@ import temple.generate.CRUD.{CRUD, presentParticiple}
...
+  private[service] def generateImports(root: ServiceRoot): String = {
+    val daoImport = s"${doubleQuote(root.module + "/dao")}"
+    if (root.projectUsesAuth)
+      mkCode("import", CodeWrap.parens.tabbed(daoImport, s"${doubleQuote(root.module + "/util")}"))
+    else mkCode("import", daoImport)
+  }
+
+  private def generateBeforeHookType(root: ServiceRoot, operation: CRUD): String = {
+    val authArg =
+      if (root.projectUsesAuth) ", auth *util.Auth"
```
jaylees14: Hmm, how about rewriting this using `.prefix` and `.parens` from CodeWrap?

jaylees14 (approved):
Nice, just one minor thing that would make it a little nicer 😄

src/main/scala/temple/generate/server/go/service/GoServiceHookGenerator.scala
willjvsmith: I knew you were gonna make this comment, the reason I didn't is because I couldn't be arsed- but I'll do it 😉

willjvsmith:
Getting you back in @jaylees14 to see how nice that refactor is due to how CodeWrap is implemented

jaylees14 (approved):
Nice <3

c-harding (approved):
Code looks good (feels like we might need some examples of hook usage soon though), not quite sure how this doesn’t help with creation though

PR #357, opened 2020-04-17
[Gen] Fix string decapitalisation with the string "ID"

c-harding:
Due to a regex typo, strings consisting of a single acronym were not decapitalising the final letter. This meant "ID" was being decapitalised to "iD", which looks painful.


Lewiky (approved):
LGTM

jaylees14 (approved):

PR #358, opened 2020-04-17
[examples] Add shopping example

jaylees14:
Add a shopping example templefile - David's idea.

Felt very similar to Amazon/AirBnb, which I guess is a good sign that everything really is the same 😄


c-harding (approved):
I will order one of everything from this shop please

examples/shopping.temple
```diff
@@ -0,0 +1,89 @@
...
+  // implicit #writable(by: this) since project has auth
+}
+
+// Customers place orders for Shoppers to deliver
+Customer: service {
+  // implicit 'id' field that matches auth service 'id' field
+  firstName: string;
+  lastName: string;
+
+  address: string;
```
Lewiky: For consistency can this be `addressLine1`?

Lewiky (approved):
Very good

PR #360, opened 2020-04-20
[Testing] [Go-gen] Fix illegal shadowing of ID in test

c-harding:
For some reason `id` was being shadowed as an int, which was leading to some illegal SQL statements being generated.

### DIFF
--- a/src/test/scala/temple/builder/BuilderTestData.scala
+++ b/src/test/scala/temple/builder/BuilderTestData.scala
@@ -1,6 +1,6 @@
 package temple.builder
 
-import temple.ast.AbstractAttribute.Attribute
+import temple.ast.AbstractAttribute.{Attribute, IDAttribute}
 import temple.ast.AbstractServiceBlock._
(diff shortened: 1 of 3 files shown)


jaylees14 (approved):
Cool. Although this test is technically valid, given the input it did provide correct output. It's just that the ID field was assumed to be the primary key.

This was also done way before we had the concept of `IDAttribute`, or even `UUID` primary keys.

c-harding:
@jaylees14 I thought that would have been the case, but I was updating it for structs and thought I may as well update it here too. Tests are often where people look to understand how our stuff works

jaylees14:
> @jaylees14 I thought that would have been the case, but I was updating it for structs and thought I may as well update it here too. Tests are often where people look to understand how our stuff works

Absolutely - wasn't a dig at all, just thought I'd give some context - thanks for fixing 😄

Lewiky (approved):
Cool, nice job

PR #363, opened 2020-04-20
[validator] Add additional constraints on foreign keys

jaylees14:
Add an additional constraint on foreign keys that prevents them from referencing a struct that isn't from their own service. 

The tests allow:
```
Foo: service {
  Bar: struct {
    foo: Foo;
  }
}
```
```
Foo: service {
  Bar: struct {}
  Baz: struct {
    bar: Bar;
  }
}
```

The tests disallow:

```
Foo: service {
  baz: Baz;
}

Bar: service {
  Baz: struct {
    ..
  }
}
```

### DIFF
--- a/src/test/scala/temple/DSL/semantics/ValidatorTest.scala
+++ b/src/test/scala/temple/DSL/semantics/ValidatorTest.scala
@@ -268,6 +268,11 @@ class ValidatorTest extends FlatSpec with Matchers {
             attributes = Map(
               "baz" -> Attribute(ForeignKey("Baz")),
             ),
+            structs = Map(
+              "Quux" -> StructBlock(
+                attributes = Map("baz" -> Attribute(ForeignKey("Baz"))),
+              ),
+            ),
           ),
           "Bar" -> ServiceBlock(
             attributes = Map(),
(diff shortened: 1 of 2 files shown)


src/test/scala/temple/DSL/semantics/ValidatorTest.scala
```diff
@@ -215,6 +215,75 @@ class ValidatorTest extends FlatSpec with Matchers {
...
+    validationErrors(
+      Templefile(
+        "MyProject",
+        projectBlock = ProjectBlock(Seq(Database.Postgres)),
+        services = Map(
+          "Foo" -> ServiceBlock(
+            attributes = Map(
+              "baz" -> Attribute(ForeignKey("Baz")),
+            ),
+          ),
```
c-harding: can we add a struct-to-foreign-struct to this test too plz?
```suggestion
          "Foo" -> ServiceBlock(
            attributes = Map(
              "baz" -> Attribute(ForeignKey("Baz")),
            ),
            structs = Map(
              "Quux" -> StructBlock(
                attributes = Map(
                  "baz" -> Attribute(StringType(ForeignKey("Baz"))),
                )
              )
            )
          ),
```

c-harding (approved):
v. nice

Lewiky (approved):

PR #364, opened 2020-04-20
[Go-Gen] Migrate to new validator library

Lewiky:
Moves from https://github.com/asaskevich/govalidator to github.com/go-playground/validator for struct validation. 

Fixes the issues with not being able to have only an upper, or lower bound on a size of something. 

A lot of test changes here, should be pretty consistent!


jaylees14 (approved):
An absolute stunner

c-harding (approved):
lgtm

PR #366, opened 2020-04-20
[go-gen] Fix issues with generation of annotated attributes

jaylees14:
This should hopefully ensure we have correct generation of each of the 3 main annotations: `@server`, `@client` and `@serverSet`.

As a summary, here's where they should/shouldn't be included:

```
/* +------------+---------+----------+-----+
   |            | Request | Response | DAO |
   +------------+---------+----------+-----+
   | @server    | N       | N        | Y   |
   +------------+---------+----------+-----+
   | @client    | Y       | N        | N   |
   +------------+---------+----------+-----+
   | @serverSet | N       | Y        | Y   |
   +------------+---------+----------+-----+
*/
```

This PR does a couple of things:
- Ensures that these attributes are generated as per that table
- Adds an additional E2E test that includes one example of each one of these attributes
  - An E2E test made the most sense, since we needed to also check OpenAPI, as well as the SQL generated for the table
- Adds an integration test for the new temple file, to check the generated code is valid
  - This involves creating a new method for building and testing an entire project, to avoid some repeated logic

Hopefully the changes to the actual `.scala` code should be pretty small: I just wanted an actual test case to make sure the entire pipeline was correct.


Lewiky (approved):
Nicely done

src/test/scala/temple/testfiles/attributes.temple
```diff
@@ -0,0 +1,24 @@
...
+/* +------------+---------+----------+-----+
+   |            | Request | Response | DAO |
+   +------------+---------+----------+-----+
+   | @server    | N       | N        | Y   |
+   +------------+---------+----------+-----+
+   | @client    | Y       | N        | N   |
+   +------------+---------+----------+-----+
+   | @serverSet | N       | Y        | Y   |
+   +------------+---------+----------+-----+
+*/
```
c-harding, line 13: Has this been added to the docs?

src/main/scala/temple/builder/ServerBuilder.scala
```diff
@@ -49,9 +49,10 @@ object ServerBuilder {
     val writable =
       serviceBlock.lookupLocalMetadata[Metadata.Writable].getOrElse(ProjectConfig.getDefaultWritable(projectUsesAuth))
 
+    val storedAttributes = serviceBlock.attributes.filter { case (_, attr) => attr.isStored }
     val queries: SortedMap[CRUD, String] =
       DatabaseBuilder
-        .buildQuery(serviceName, serviceBlock.attributes, endpoints, readable, selectionAttribute = idAttribute.name)
+        .buildQuery(serviceName, storedAttributes, endpoints, readable, selectionAttribute = idAttribute.name)
```
c-harding: `serviceBlock.storedAttributes`?

src/it/scala/temple/containers/GolangSpec.scala
```diff
@@ -26,4 +32,23 @@ abstract class GolangSpec extends DockerShell2HttpService(8081) with DockerTestK
...
+    analyzedTemplefile.allServices.keys.map(StringUtils.kebabCase).map { service =>
+      val filesInService = project.files.filter { case (file, _) => file.folder.startsWith(service) }
+      val validationErrors = validateAll(
+        filesInService,
+        entryFile = File(service, s"$service.go"),
+      )
+
+      validationErrors
+    }
+  }
```
c-harding, line 53: Where’s this come from?

c-harding (approved):
very nice

src/test/scala/temple/testfiles/attributes.temple
jaylees14: It will be

src/it/scala/temple/containers/GolangSpec.scala
jaylees14: It was previously part of the single `simple.temple` test, but we are reusing this logic of building from templefile in a second test, so I've extracted out.

PR #367, opened 2020-04-20
[openapi-gen] Add auth service to OpenAPI gen

jaylees14:
Adds the auth service to OpenAPI generator. 

I feel like the `Auth` trait should be moved somewhere else? Thoughts?

### DIFF
--- a/src/main/scala/temple/generate/target/openapi/OpenAPIGenerator.scala
+++ b/src/main/scala/temple/generate/target/openapi/OpenAPIGenerator.scala
@@ -22,6 +22,7 @@ private class OpenAPIGenerator private (name: String, version: String, descripti
   private val errorTracker = FlagMapView(
     400 -> generateError("Invalid request", "Invalid request parameters: name"),
     401 -> generateError("Valid request but forbidden by server", "Not authorised to create this object"),
+    403 -> generateError("Valid request but server will not fulfill", "User with this ID already exists"),
     404 -> generateError("ID not found", "Object not found with ID 1"),
     500 -> generateError(
       "The server encountered an error while serving this request",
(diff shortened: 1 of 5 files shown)


src/main/scala/temple/generate/target/openapi/OpenAPIGenerator.scala
```diff
@@ -22,6 +22,7 @@ private class OpenAPIGenerator private (name: String, version: String, descripti
   private val errorTracker = FlagMapView(
     400 -> generateError("Invalid request", "Invalid request parameters: name"),
     401 -> generateError("Valid request but forbidden by server", "Not authorised to create this object"),
+    403 -> generateError("Valid request but server will not fulfill", "User with this ID already exists"),
```
willjvsmith, line 25: This made me think, what if the user adds endpoints in their service, then updates the openAPI doc. It gets overwritten when regenerating 😱 do we need hooks for openAPI?!!!

willjvsmith (approved):
LGTM 🚀 

Forgive me for being dense, but could you not continue use the `AuthMethod` from `Metadata`. I'm not familiar with the architecture being used to generate these yaml files.

jaylees14:
> LGTM 🚀 
> 
> 
> 
> Forgive me for being dense, but could you not continue use the `AuthMethod` from `Metadata`. I'm not familiar with the architecture being used to generate these yaml files.

We could do, but we usually keep the generators separate from the AST so there is some level of decoupling. They usually have their own AST which gets mapped into in `ProjectBuilder`

src/main/scala/temple/generate/target/openapi/OpenAPIGenerator.scala
jaylees14: Good point.. no idea how we’d do that 😂

willjvsmith:
> We could do, but we usually keep the generators separate from the AST so there is some level of decoupling. They usually have their own AST which gets mapped into in ProjectBuilder

I see, I reused AST types in the Go generation if they were exactly the same

Lewiky (approved):
Nicely done

PR #368, opened 2020-04-20
[openapi-gen] Fix service name in paths

jaylees14:
Just a little one: kebab case the url names to match the URL defined in Go.


Lewiky (approved):
Yup, nice and easy

c-harding (approved):

PR #371, opened 2020-04-21
[go-gen] Infer which endpoints to generate

c-harding:
Don’t specify which endpoints to generate, as this can fully be inferred from the templefile being passed in

PR #372, opened 2020-04-21
[Go-gen] Add struct-root to generate from

c-harding:
On top of #371 (to simplify logic of generating endpoints, by only having to do that in one place), this now provides a StructRoot, equivalent to a ServiceRoot, to be used by the go-gen to generate struct code


jaylees14 (approved):

examples/airbnb.temple
```diff
@@ -37,7 +37,6 @@ Property: service {
   Photo: struct {
     // implicit 'id' field
     // implicit 'parentID' field
-    // implicit 'createdBy' field
```
willjvsmith, line 40: Might need to update the shopping example too

willjvsmith (approved):
LGTM 🚀

examples/airbnb.temple
c-harding: Can’t see any incorrect comments in shopping

PR #373, opened 2020-04-21
[SQL-gen] Generate queries for structs

c-harding:
Following from #372, with the TODO done: here are the SQL queries for interfacing with a struct in the database

### DIFF
--- a/src/main/scala/temple/builder/ServerBuilder.scala
+++ b/src/main/scala/temple/builder/ServerBuilder.scala
@@ -11,11 +11,11 @@ import temple.detail.LanguageDetail
 import temple.detail.LanguageDetail.GoLanguageDetail
 import temple.generate.CRUD._
 import temple.generate.database.{PostgresContext, PostgresGenerator, PreparedType}
-import temple.generate.server
 import temple.generate.server.AttributesRoot.{ServiceRoot, StructRoot}
 import temple.generate.server._
 import temple.utils.StringUtils
 
+import scala.Option.when
 import scala.collection.immutable.{ListMap, SortedMap}
 
 object ServerBuilder {
@@ -36,16 +36,14 @@ object ServerBuilder {
     // Whether or not this service has an auth block
     val hasAuthBlock = serviceBlock.lookupLocalMetadata[ServiceAuth].isDefined
 
-    // The created by attribute exists if the project has auth and this service does not have an auth block
-    val createdBy =
-      if (projectUsesAuth && !hasAuthBlock)
-        Some(CreatedByAttribute(languageConfig.createdByInputName, languageConfig.createdByName))
-      else None
+    val createdBy = when(serviceBlock.attributes.valuesIterator contains AbstractAttribute.CreatedByAttribute) {
+      CreatedByAttribute(languageConfig.createdByInputName, languageConfig.createdByName)
+    }
 
-    val idAttribute = server.IDAttribute("id")
+    val idAttribute = IDAttribute("id")
 
     val readable =
-      serviceBlock.lookupLocalMetadata[Metadata.Readable].getOrElse(ProjectConfig.getDefaultReadable(projectUsesAuth))
+      serviceBlock.lookupLocalMetadata[Metadata.Readable] getOrElse ProjectConfig.getDefaultReadable(projectUsesAuth)
     val writable =
       serviceBlock.lookupLocalMetadata[Metadata.Writable] getOrElse ProjectConfig.getDefaultWritable(projectUsesAuth)
 
@@ -53,6 +51,7 @@ object ServerBuilder {
       serviceName,
       serviceBlock.storedAttributes,
       endpoints(serviceBlock),
+      isStruct = false,
       idAttribute.name,
       database,
       readable,
@@ -77,9 +76,19 @@ object ServerBuilder {
       case (structName: String, structBlock: StructBlock) =>
         val idAttribute = IDAttribute("id")
 
+        val queries = buildQueries(
+          structName,
+          structBlock.attributes,
+          endpoints(structBlock),
+          isStruct = true,
+          idAttribute.name,
+          database,
+          readable,
+        )
+
         StructRoot(
           name = structName,
-          opQueries = SortedMap(),
+          opQueries = queries,
           idAttribute = idAttribute,
           createdByAttribute = None,
           parentAttribute = Some(ParentAttribute("parentID")),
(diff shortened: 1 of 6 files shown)


src/test/scala/temple/builder/ServerBuilderTest.scala
```diff
@@ -108,7 +108,10 @@ class ServerBuilderTest extends FlatSpec with Matchers {
         StructRoot(
           name = "Test",
           opQueries = SortedMap(
-            // TODO: Coming soon
+            Create -> "INSERT INTO test (favourite_colour, bed_time, favourite_number) VALUES ($1, $2, $3) RETURNING favourite_colour, bed_time, favourite_number;",
+            Read   -> "SELECT favourite_colour, bed_time, favourite_number FROM test WHERE id = $1;",
+            Update -> "UPDATE test SET favourite_colour = $1, bed_time = $2, favourite_number = $3 WHERE id = $4 RETURNING favourite_colour, bed_time, favourite_number;",
+            Delete -> "DELETE FROM test WHERE id = $1;",
```
willjvsmith, line 114: I don't think we've included trailing semi-colons elsewhere

src/main/scala/temple/builder/ServerBuilder.scala
```diff
@@ -27,34 +27,34 @@ object ServerBuilder {
     detail: LanguageDetail,
     projectUsesAuth: Boolean,
   ): ServiceRoot = {
-    val language = serviceBlock.lookupMetadata[ServiceLanguage].getOrElse(ProjectConfig.defaultLanguage)
-    val database = serviceBlock.lookupMetadata[Database].getOrElse(ProjectConfig.defaultDatabase)
+    val language = serviceBlock.lookupMetadata[ServiceLanguage] getOrElse ProjectConfig.defaultLanguage
+    val database = serviceBlock.lookupMetadata[Database] getOrElse ProjectConfig.defaultDatabase
```
willjvsmith: why?

src/main/scala/temple/builder/ServerBuilder.scala
```diff
@@ -27,34 +27,34 @@ object ServerBuilder {
...
     val languageConfig = language match {
       case ServiceLanguage.Go => GoLanguageConfig
     }
 
     // Whether or not this service has an auth block
     val hasAuthBlock = serviceBlock.lookupLocalMetadata[ServiceAuth].isDefined
 
-    // The created by attribute exists if the project has auth and this service does not have an auth block
-    val createdBy =
-      if (projectUsesAuth && !hasAuthBlock)
```
jaylees14, line 41: Do we not need this part of the condition?

jaylees14 (approved):

src/main/scala/temple/builder/ServerBuilder.scala
c-harding: @jaylees14 this is calculated when we add a CreatedByAttribute, so it’s already been done (and the condition is slightly more complex now: struct blocks don’t have createdBy)

src/main/scala/temple/builder/ServerBuilder.scala
```diff
@@ -27,34 +27,34 @@ object ServerBuilder {
...
-        Some(CreatedByAttribute(languageConfig.createdByInputName, languageConfig.createdByName))
-      else None
+    val createdBy = when(serviceBlock.attributes.valuesIterator contains AbstractAttribute.CreatedByAttribute) {
+      CreatedByAttribute(languageConfig.createdByInputName, languageConfig.createdByName)
+    }
```
willjvsmith, line 46: Is this equivalent?

willjvsmith (approved):
LGTM 🚀 

You've done some refactoring here that isn't mentioned in the description

src/test/scala/temple/builder/ServerBuilderTest.scala
c-harding: @smithwjv this is from mkCode.stmt, see the existing queries in https://github.com/TempleEight/temple/blob/develop/src/test/scala/temple/builder/ServerBuilderTest.scala

src/main/scala/temple/builder/ServerBuilder.scala
c-harding: I was moving the two lines from below (`readable`, `writable`) up, and one of them was infix already, so I made them consistent. Happy to switch them all to methods if you’d prefer

src/main/scala/temple/builder/ServerBuilder.scala
c-harding: @smithwjv `when` is equivalent to the if-then-some-else-none. The condition has now been moved to the [validator](https://github.com/TempleEight/temple/blob/730a8c27462c98346f766e73cbfb9b4c176609a0/src/main/scala/temple/DSL/semantics/Validator.scala#L37), so it’s calculated in just one place rather than both, but otherwise the same

src/test/scala/temple/builder/DatabaseBuilderTest.scala
```diff
@@ -90,9 +95,12 @@ class DatabaseBuilderTest extends FlatSpec with Matchers {
...
       BuilderTestData.sampleService.attributes,
       Set(CRUD.List),
+      isStruct = false,
       Metadata.Readable.This,
     )
     queries.keys should contain(CRUD.List)
     queries(CRUD.List) shouldBe DatabaseBuilderTestData.sampleListStatementEnumerateByCreator
   }
+
+  // TODO: test the list endpoint with isStruct=true
```
jaylees14: This would actually be quite good to do since it tests the disjunction logic. Could you add one please

src/main/scala/temple/builder/ServerBuilder.scala
willjvsmith: I think we've avoided this notation elsewhere that isn't tests

src/main/scala/temple/builder/ServerBuilder.scala
willjvsmith: Yeah I was referring to the condition, I see, fair enough

src/test/scala/temple/builder/DatabaseBuilderTest.scala
c-harding: I’ve realised this will never happen anyway, and so removed that disjunction function: a struct can never have createdBy, so it’s one or the other, but not both

src/main/scala/temple/builder/ServerBuilder.scala
c-harding: fixed

PR #375, opened 2020-04-21
[SQL-gen] Add parent attribute to SQL database

c-harding:
On top of #373. Every struct will have an implicit `parentID` stored.


Lewiky (approved):
Awesome

jaylees14 (approved):

jaylees14 (approved):

PR #376, opened 2020-04-21
[Go-gen] Generate the DAO for the structs

c-harding:
Here’s the first of the actual Go generation: the DAO.

Scala changes: +171-160

In the DAO, we now generate DatastoreObjectStruct, Structs, DAOFunctions for the structs as well as the services, and we generate QueryFunctions for every allowed operation, whether that operation is performed on the service or the struct.

A lot of `root` is renamed to `block`, as it is generalised from ServiceRoot to AttributesRoot.

`enumeratingByCreator` is dropped as a parameter, as it is was getting everywhere, and it’s calculable from the block itself.

### DIFF
--- a/src/e2e/resources/simple-temple-expected/simple-temple-test-user/dao/dao.go
+++ b/src/e2e/resources/simple-temple-expected/simple-temple-test-user/dao/dao.go
@@ -273,7 +273,7 @@ func (dao *DAO) ListFred(input ListFredInput) (*[]Fred, error) {
 
 // CreateFred creates a new fred in the datastore, returning the newly created fred
 func (dao *DAO) CreateFred(input CreateFredInput) (*Fred, error) {
-	row := executeQueryWithRowResponse(dao.DB, "INSERT INTO fred (id, parent_id, field, friend, image) VALUES ($1, $2, $3, $4, $5) RETURNING id, parent_id, field, friend, image;", input.ID, input.Field, input.Friend, input.Image)
+	row := executeQueryWithRowResponse(dao.DB, "INSERT INTO fred (id, parent_id, field, friend, image) VALUES ($1, $2, $3, $4, $5) RETURNING id, parent_id, field, friend, image;", input.ID, input.ParentID, input.Field, input.Friend, input.Image)
 
 	var fred Fred
 	err := row.Scan(&fred.ID, &fred.Field, &fred.Friend, &fred.Image)
--- a/src/e2e/resources/simple-temple-expected/simple-temple-test-user/dao/dao.go
+++ b/src/e2e/resources/simple-temple-expected/simple-temple-test-user/dao/dao.go
@@ -249,7 +249,7 @@ func (dao *DAO) IdentifySimpleTempleTestUser(input IdentifySimpleTempleTestUserI
 
 // ListFred returns a list containing every fred in the datastore
 func (dao *DAO) ListFred(input ListFredInput) (*[]Fred, error) {
-	rows, err := executeQueryWithRowResponses(dao.DB, "SELECT id, created_by, field, friend, image FROM fred WHERE parent_id = $1;", input.ParentID)
+	rows, err := executeQueryWithRowResponses(dao.DB, "SELECT id, parent_id, field, friend, image FROM fred WHERE parent_id = $1;", input.ParentID)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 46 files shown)


src/e2e/resources/simple-temple-expected/simple-temple-test-user/dao/dao.go
```diff
@@ -191,3 +246,88 @@ func (dao *DAO) IdentifySimpleTempleTestUser(input IdentifySimpleTempleTestUserI
...
+func (dao *DAO) ListFred(input ListFredInput) (*[]Fred, error) {
+	rows, err := executeQueryWithRowResponses(dao.DB, "SELECT id, created_by, field, friend, image FROM fred WHERE parent_id = $1;", input.ParentID)
+	if err != nil {
+		return nil, err
+	}
+
+	fredList := make([]Fred, 0)
+	for rows.Next() {
+		var fred Fred
+		err = rows.Scan(&fred.ID, &fred.Field, &fred.Friend, &fred.Image)
```
jaylees14: We select `created_by`, but then don't scan the value into the struct. This works right?

src/e2e/resources/simple-temple-expected/simple-temple-test-user/dao/dao.go
```diff
@@ -191,3 +246,88 @@ func (dao *DAO) IdentifySimpleTempleTestUser(input IdentifySimpleTempleTestUserI
...
+	if err != nil {
+		return nil, err
+	}
+
+	return &fredList, nil
+}
+
+// CreateFred creates a new fred in the datastore, returning the newly created fred
+func (dao *DAO) CreateFred(input CreateFredInput) (*Fred, error) {
+	row := executeQueryWithRowResponse(dao.DB, "INSERT INTO fred (id, created_by, field, friend, image) VALUES ($1, $2, $3, $4, $5) RETURNING id, created_by, field, friend, image;", input.ID, input.Field, input.Friend, input.Image)
```
jaylees14: Where do we set the `parent_id`?

Lewiky (approved):
block block block

Lewiky (approved):
block block block

Lewiky (approved):
block block block

src/main/scala/temple/generate/server/go/service/dao/GoServiceDAOInterfaceGenerator.scala
```diff
@@ -1,44 +1,54 @@
...
 
-  private[service] def generateInterface(
-    root: ServiceRoot,
-    enumeratingByCreator: Boolean,
-  ): String =
+  private[service] def generateInterface(root: ServiceRoot): String =
     mkCode.lines(
       "// BaseDatastore provides the basic datastore methods",
       mkCode(
         "type BaseDatastore interface",
```
jaylees14, line 41: Isn't there a utility method for creating interfaces somewhere?

Lewiky:
BLOCKOOOOCK

jaylees14 (commented):
Cool, if a little big. Might have been easier to see the changes if you changed the type to attribute root first, and then include the structs in the output? Just a suggestion.

I'm also curious if you've run the DAO code. Where does `parent_id` get set?

Lewiky:
BLOCKOOOOCK

src/e2e/resources/simple-temple-expected/simple-temple-test-user/dao/dao.go
c-harding: Aah, I hadn’t run this without #375 (I rebased it so it could be reviewed sooner), forgetting that the sql updates would copy through to here. Fixed that now. I’ve run it, but in this commit it doesn’t get run yet because the calling of these hooks hasn’t yet happened.

src/e2e/resources/simple-temple-expected/simple-temple-test-user/dao/dao.go
c-harding: should do here, now that’s been merged in

src/main/scala/temple/generate/server/go/service/dao/GoServiceDAOInterfaceGenerator.scala
c-harding: not that I know of, but there may be. I think this was written by @smithwjv, so he may know why one wasn’t used

src/test/resources/project-builder-complex/complex-user/dao/dao.go
```diff
@@ -194,3 +250,64 @@ func (dao *DAO) IdentifyComplexUser(input IdentifyComplexUserInput) (*ComplexUse
 
 	return &complexUser, nil
 }
+
+// CreateTempleUser creates a new templeUser in the datastore, returning the newly created templeUser
+func (dao *DAO) CreateTempleUser(input CreateTempleUserInput) (*TempleUser, error) {
+	row := executeQueryWithRowResponse(dao.DB, "INSERT INTO temple_user (id, int_field, double_field, string_field, bool_field, date_field, time_field, date_time_field, blob_field) VALUES ($1, $2, $3, $4, $5, $6, $7, $8, $9) RETURNING id, int_field, double_field, string_field, bool_field, date_field, time_field, date_time_field, blob_field;", input.ID, input.ParentID, input.IntField, input.DoubleField, input.StringField, input.BoolField, input.DateField, input.TimeField, input.DateTimeField, input.BlobField)
```
jaylees14: ParentID isn't in this query?

jaylees14 (commented):
Hmm still not fully sold, where's the `parentID` in the `TempleUser` struct?

src/test/resources/project-builder-complex/complex-user/dao/dao.go
c-harding: Ffs, I’ve been running this from templefiles, but this code is generated from a unit test: the input is post-validate stage, and wasn’t valid. Any idea how to prevent cases like that? Just manually checking every unit test every time the validator is updated?

jaylees14: Hmm not really too sure. I guess this is an issue with where we inject the `ParentID` attribute. It feels as though it shouldn't be added in the validator stage, but rather it's implicit from the actual thing you're trying to generate. That way it's not possible to construct an invalid object to generate using the available constructors.

jaylees14 (approved):

PR #377, opened 2020-04-21
[fix][regen] Add buildFilenameSeq function

### LINKED ISSUE #374: Regeneration filter crashes on hidden files
To reproduce:

* `mkdir example`
* `temple generate -o example path/to/templefile`

If the `example` directory has a hidden file, such as `.DS_Store`, the regeneration filter crashes

Lewiky:
Adds a `buildFilenameSeq` function to `FileUtils` which mirrors `buildFileMap`, except it just returns a sequence of the `File` objects found rather than their contents.

Closes #374


src/main/scala/temple/utils/FileUtils.scala
```diff
@@ -105,4 +112,36 @@ object FileUtils {
...
     foundFiles
   }
+
+  /**
+    * Build a sequence of filenames of files present in the provided filepath
+    * @param path The root path to explore
+    * @param cwd The location to give paths relative to
+    * @return A sequence of filenames, relative to the provided path
+    */
+  private def buildFilenameMap(path: Path, cwd: Path): Seq[File] = {
```
c-harding: is this a `Map`? Odd name for the function

src/main/scala/temple/utils/FileUtils.scala
```diff
@@ -105,4 +112,36 @@ object FileUtils {
...
+    var foundFiles: Seq[File] = Seq.empty
```
c-harding: I realise it’s the same for the map, but this would be better served by a val of type `mutable.Buffer`, to avoid all the list copying.

src/main/scala/temple/utils/FileUtils.scala
```diff
@@ -105,4 +112,36 @@ object FileUtils {
...
+        }
+
+        override def visitFileFailed(file: Path, exc: IOException): FileVisitResult =
+          throw exc
+
+        override def postVisitDirectory(dir: Path, exc: IOException): FileVisitResult =
+          FileVisitResult.CONTINUE
+      },
+    )
+    foundFiles
```
c-harding: (and then foundFiles.toSeq here)

c-harding (approved):
nice

src/main/scala/temple/utils/FileUtils.scala
Lewiky: Whoops renamed the other but not this one

jaylees14 (approved):

PR #378, opened 2020-04-21
[go-gen] Name hooks after their service

c-harding:
This is in preparation for introducing hooks for structs.

No stacking, so ready for review immediately.

A bit more generalising `root: ServiceRoot` to `block: AttributesBlock`, and otherwise basically all the changes are of test resources

### DIFF
--- a/src/e2e/resources/attributes-temple-expected/example/example.go
+++ b/src/e2e/resources/attributes-temple-expected/example/example.go
@@ -127,7 +127,7 @@ func (env *env) createExampleHandler(w http.ResponseWriter, r *http.Request) {
 		ID: uuid,
 	}
 
-	for _, hook := range env.hook.beforeCreateHooks {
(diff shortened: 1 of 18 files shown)


jaylees14 (approved):
Ok. This wasn't done in `spec-golang`, nor was it discussed anywhere, so why the change?

c-harding:
This was done to increase consistency between structs and services, so they wouldn’t clash, and it makes the code gen easier without reducing readability. Do you think it would be better to leave the service hook names shorter than the struct hook names?

willjvsmith (changes requested):
I think the hooks should remain as they are, and the struct ones have the suffix, like `beforeCreatePictureHook`. Since structs are an addition to the service, the service itself is the base.

As @jaylees14 says, I think it would be wise to discuss changes like this before making them

PR #379, opened 2020-04-21
[hotfix][analyser] Fix min and max int arguments being in the wrong order

Lewiky:
…rder


c-harding (approved):
nice, thanks

jaylees14 (approved):

willjvsmith (approved):
LGTM 🚀 

Do we not use these in any of the other tests?

PR #380, opened 2020-04-21
[test] Fix flaky integration test

jaylees14:
- Bump version of Kong to 2.0.2
- Rather than testing if a service is accessible, determine if Kong has the routes registered
- Re-add the additional sleep that was used to allow the routes to propagate to Kong before any requests are made


c-harding (approved):
thank you <3

willjvsmith (approved):
LGTM 🚀

PR #381, opened 2020-04-21
[Go-gen] Add struct hooks without renaming old hooks

c-harding:
After #378 was shouted down, here’s a second attempt at adding hooks for structs. Includes more renaming `root: ServiceRoot` to `block: AttributesRoot`: I’m doing this as I widen the types of each function. I can’t really do that refactor separately, as it means the function now needs to be able to handle StructBlocks too

The fall in codecov is a single line added to keep a match exhaustive: it throws an explanatory error when an impossible case is reached.


jaylees14 (approved):
Nice 👍

Lewiky (approved):
LGTM

PR #382, opened 2020-04-21
[Testing] [go-gen] Print go directory structure on Go compilation error

c-harding:
This makes dealing with Go compilation errors slightly easier. The line numbers are printed with the source code.

Also simplifies the logic for encoding `Files` to JSON, using circe’s KeyEncoder

### DIFF
--- a/src/it/scala/temple/containers/GolangSpec.scala
+++ b/src/it/scala/temple/containers/GolangSpec.scala
@@ -5,17 +5,16 @@ import com.whisk.docker.DockerFactory
 import com.whisk.docker.impl.spotify.SpotifyDockerFactory
 import com.whisk.docker.scalatest.DockerTestKit
 import io.circe.syntax._
-import org.scalatest.BeforeAndAfterAll
 import scalaj.http.Http
 import temple.DSL.DSLProcessor
 import temple.DSL.semantics.Analyzer
 import temple.builder.project.ProjectBuilder
 import temple.detail.LanguageDetail.GoLanguageDetail
 import temple.generate.FileSystem._
-import temple.utils.StringUtils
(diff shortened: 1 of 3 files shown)


src/it/scala/temple/containers/GolangSpec.scala
```diff
@@ -25,12 +26,21 @@ abstract class GolangSpec extends DockerShell2HttpService(8081) with DockerTestK
...
       .params(Map("src" -> json, "root" -> entryFile.folder, "entrypoint" -> entryFile.filename))
-      .timeout(connTimeoutMs = 1000, readTimeoutMs = 30000)
+      .timeout(connTimeoutMs = 1000, readTimeoutMs = 60000)
       .asString
       .body
+    if (errors.nonEmpty) {
+      for ((file, content) <- files.to(SortedMap)) {
+        System.err.println(s"$file:")
+        for ((line, i) <- content.linesIterator.zipWithIndex) {
+          System.err.println(s"${i + 1}:".padTo(6, ' ') + line)
```
jaylees14: 6? 🧙

jaylees14 (approved):
Nice

src/it/scala/temple/containers/GolangSpec.scala
c-harding: surely no more than 4 digits for the row number, usually 3, then a colon, then a space. Would you rather I commented it?

jaylees14: I don’t think I would have guessed that explanation, so yes please 👍

jaylees14 (approved):

Lewiky (approved):
Nice one

PR #383, opened 2020-04-21
[Go-gen] Add method for getting parent ID from a URL

c-harding:
Generate a method for extracting the parent ID from a URL, and output it whenever a service has structs.

Side effects: add a method for determining whether an AttributesRoot describes a struct

PR #384, opened 2020-04-21
[Go-gen] Add metrics entries for structs

c-harding:
On top of #383, so the diff will shrink shortly. Preview the diff [here](https://github.com/TempleEight/temple/pull/384/commits/dd39215110baaf57be682abe278b50db06293d08), it’s only +39-13.

Add the strings for use with metric services to `metric.go`. Still to do is to add them to the dashboard.


Lewiky (approved):
Easy peasy

jaylees14 (approved):
Nice

PR #385, opened 2020-04-22
[Go-gen] Struct handlers

c-harding:
Here come the structs! Adds functions for checking parentage, the handler functions, and the kong registration functions.

Once #392 is merged, it’s +1,172 -56, which is +219 -56 of code plus 953 new lines of Go code (all very repetitive and basically the same as what would be generated if the structs were duplicated).

Currently failing the Deliveroo integration test due to an existing issue, as shown in #386.


Lewiky (approved):
Alright cool cool

src/main/scala/temple/generate/server/go/service/main/GoServiceMainDeleteHandlerGenerator.scala
```diff
@@ -19,7 +20,11 @@ object GoServiceMainDeleteHandlerGenerator {
       "input",
     )
 
-  private def generateDAOCallBlock(block: AttributesRoot, metricSuffix: Option[String]): String =
+  private def generateDAOCallBlock(
+    block: AttributesRoot,
+    parent: Option[ServiceRoot],
```
jaylees14: Is this param used?

src/main/scala/temple/generate/server/go/service/main/GoServiceMainReadHandlerGenerator.scala
```diff
@@ -3,17 +3,23 @@ package temple.generate.server.go.service.main
...
 
 import scala.Option.when
 import scala.collection.immutable.ListMap
 
 object GoServiceMainReadHandlerGenerator {
 
-  private def generateDAOCallBlock(block: AttributesRoot, metricSuffix: Option[String]): String =
+  private def generateDAOCallBlock(
+    block: AttributesRoot,
+    parent: Option[ServiceRoot],
```
jaylees14: Ditto here?

jaylees14 (approved):
Cool, much easier to review when it's just the additions. Generated code looks good, would normally do some manual testing but I'll save that for `temple test`

PR #386, opened 2020-04-22
[example] Add ImageServer example

c-harding:
This is going to break because an attribute (of type data or date/time) having the same name as its class will result in a name clash (source: https://github.com/TempleEight/temple/blob/develop/src/main/scala/temple/generate/server/go/service/main/GoServiceMainCreateHandlerGenerator.scala#L90). I’d be interested in people’s solutions to this: ban all attributes from having the same name as their services, rename the new struct in the create block from `image` to `newImage`, or ban specifically this case?

I raise this because the Deliveroo example has the equivalent of this in a struct, which my generation has raised.

Update: now naming an attribute the same as its block is a validation error. This triggered a naming error in the simple.temple, which has been updated at the cost of many lines.

Also spotted was another issue based on the original incorrect assumption: attributes may not be named the same as imports etc, as they are sometimes rendered as local variables. This has been fixed too, with the changes to NameClashes

_Previous solution: rename the parent in the case of a clash. This means the service is now called ImageServerImage, and the attribute remains as before._ (issue: the API should not change that much when children are added.)

### DIFF
--- a/examples/airbnb.temple
+++ b/examples/airbnb.temple
@@ -70,7 +70,7 @@ Review: service {
   tenant: Tenant;
   reservation: Reservation;
   stars: int(5, 0);
-  review: string;
+  message: string;
   #readable(by: all);
   // implicit #writable(by: this) since project has auth
 }
--- a/examples/deliveroo.temple
+++ b/examples/deliveroo.temple
@@ -69,7 +69,7 @@ Dish: service {
     restaurant: Restaurant;
 
     Img: struct {
-        img: data(4M);
+        data: data(4M);
         #enumerable;
     }
     #enumerable;
--- a/examples/image-server.temple
+++ b/examples/image-server.temple
@@ -7,7 +7,7 @@ ImageServer: project {
 }
 
 Image: service {
-    image: data(4M);
+    imageData: data(4M);
     description: string;
     owner: User;
     #enumerable;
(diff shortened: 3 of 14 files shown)


jaylees14:
If we rename the variable in the Go code, we’re still going to have to ban that new name from occurring in the attributes, so I’d be tempted to say the quickest fix is banning attributes with the same name as the service.

c-harding:
actually, tempted to go with Jay’s solution even though it does interfere more with the user’s workflow: this will make regen more awkward when such an attribute is added, as the whole project is moved. Alternatively, could rename the attr?

TODO: also check that the current approach happens after the attr has been renamed

Lewiky:
I think going with Jay's solution here will be fine, The user is like 3 people and I feel like this level of detail isn't going to bother them. Not like this is going out to millions

c-harding:
@jaylees14 this is going to get messy: simple.temple has `User: service { user: string; ... }`

jaylees14 (approved):
Cool

Lewiky (approved):
coolio

c-harding:
Update: now naming an attribute the same as its block is a validation error. This triggered a naming error in the simple.temple, which has been updated at the cost of many lines.

Also spotted was another issue based on the original incorrect assumption: attributes may not be named the same as imports etc, as they are sometimes rendered as local variables. This has been fixed too, with the changes to NameClashes

PR #347, opened 2020-04-17
[DSL] Remove access controls from structs

c-harding:
As agreed on slack, this can now be inherited in its entirety from the parent service. Surprised at how easy this was to remove…

Edit: reverts #344, d’oh

PR #388, opened 2020-04-24
[Validator] Remove fruitless struct access metadata checks

c-harding:
Fixes a warning introduced in #347, by removing `#readable`/`#writable` from structs but not from a pattern match


jaylees14 (approved):
## _FRUITLESS_

Lewiky (approved):
LGTM

PR #391, opened 2020-04-25
Fix unique implementation

### LINKED ISSUE #387: Unique errors are not checked in the DAO response
From @Lewiky’s docs:

```bash
 ~/Documents/temple-tutorial ❯❯❯ curl -X POST $KONG_ENTRY/api/example-service -d '{"foo": "value!", "bar": 10}'
 {"id":"e24e6d58-83e7-11ea-b435-0242ac1d0003","foo":"value!","bar":10}

 ~/Documents/temple-tutorial ❯❯❯ curl -X POST $KONG_ENTRY/api/example-service -d '{"foo": "another value!", "bar": 10}'
 {"error":"Something went wrong: pq: duplicate key value violates unique constraint \"example_service_bar_key\""}
```

We should be returning a much nicer error than this, without exposing the pq internals

jaylees14:
Currently if we used an `@unique` attribute, the error returned would be code 500, with response:
```
{"error":"Something went wrong: pq: duplicate key value violates unique constraint \"example_service_bar_key\""}
```

This isn't ideal, since it's not an internal server error. Now we detect if this is present and return an appropriate error and code.

Closes #387.

### Test Plan
Consider the tutorial templefile:

```
ExampleProject: project {
  #provider(dockerCompose);
}

ExampleService: service {
  foo: string;
  bar: int @unique;
}
```

Let's spin it up and make some requests:

```bash
# Make a new entity
❯❯❯ curl $KONG_ENTRY/api/example-service -d '{"foo": "abcd", "bar": 123}'
{"id":"2c0d13f0-8723-11ea-90d0-0242ac1c0002","foo":"abcd","bar":123}

# Repeat Request
❯❯❯ curl $KONG_ENTRY/api/example-service -d '{"foo": "abcd", "bar": 123}'
{"error":"Duplicate ExampleService found"}
```
```bash
# Make a new entity
❯❯❯ curl $KONG_ENTRY/api/example-service -d '{"foo": "abcd", "bar": 12356}'
{"id":"35ef8443-8723-11ea-90d0-0242ac1c0002","foo":"abcd","bar":12356}

# Update the second entity with value of bar that matches the first
❯❯❯ curl -X PUT $KONG_ENTRY/api/example-service/35ef8443-8723-11ea-90d0-0242ac1c0002 -d '{"foo": "abcd", "bar": 123}'
{"error":"Duplicate ExampleService found"}
```

### DIFF
--- a/src/e2e/resources/simple-temple-expected/simple-temple-test-user/dao/dao.go
+++ b/src/e2e/resources/simple-temple-expected/simple-temple-test-user/dao/dao.go
@@ -9,7 +9,7 @@ import (
 	"github.com/google/uuid"
 
 	// pq acts as the driver for SQL requests
-	_ "github.com/lib/pq"
+	"github.com/lib/pq"
 )
 
 // BaseDatastore provides the basic datastore methods
@@ -128,6 +128,9 @@ type DeleteFredInput struct {
 	ID uuid.UUID
 }
 
+// https://www.postgresql.org/docs/9.3/errcodes-appendix.html
+const psqlUniqueViolation = "unique_violation"
+
 // Init opens the datastore connection, returning a DAO
 func Init(config *util.Config) (*DAO, error) {
 	connStr := fmt.Sprintf("user=%s dbname=%s host=%s sslmode=%s", config.User, config.DBName, config.Host, config.SSLMode)
@@ -188,6 +191,13 @@ func (dao *DAO) CreateSimpleTempleTestUser(input CreateSimpleTempleTestUserInput
 	var simpleTempleTestUser SimpleTempleTestUser
 	err := row.Scan(&simpleTempleTestUser.ID, &simpleTempleTestUser.SimpleTempleTestUser, &simpleTempleTestUser.Email, &simpleTempleTestUser.FirstName, &simpleTempleTestUser.LastName, &simpleTempleTestUser.CreatedAt, &simpleTempleTestUser.NumberOfDogs, &simpleTempleTestUser.Yeets, &simpleTempleTestUser.CurrentBankBalance, &simpleTempleTestUser.BirthDate, &simpleTempleTestUser.BreakfastTime)
 	if err != nil {
+		// pq specific error
+		if err, ok := err.(*pq.Error); ok {
+			if err.Code.Name() == psqlUniqueViolation {
+				return nil, ErrDuplicateSimpleTempleTestUser
+			}
+		}
+
 		return nil, err
 	}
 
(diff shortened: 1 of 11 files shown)


src/e2e/resources/simple-temple-expected/simple-temple-test-user/dao/dao.go
```diff
@@ -188,6 +191,13 @@ func (dao *DAO) CreateSimpleTempleTestUser(input CreateSimpleTempleTestUserInput
 	var simpleTempleTestUser SimpleTempleTestUser
 	err := row.Scan(&simpleTempleTestUser.ID, &simpleTempleTestUser.SimpleTempleTestUser, &simpleTempleTestUser.Email, &simpleTempleTestUser.FirstName, &simpleTempleTestUser.LastName, &simpleTempleTestUser.CreatedAt, &simpleTempleTestUser.NumberOfDogs, &simpleTempleTestUser.Yeets, &simpleTempleTestUser.CurrentBankBalance, &simpleTempleTestUser.BirthDate, &simpleTempleTestUser.BreakfastTime)
 	if err != nil {
+		// pq specific error
+		if err, ok := err.(*pq.Error); ok {
+			if err.Code.Name() == psqlUniqueViolation {
+				return nil, ErrDuplicateSimpleTempleTestUser("")
```
c-harding: Why the empty string?

c-harding (approved):
Nice!

src/main/scala/temple/generate/server/go/service/main/GoServiceMainCreateHandlerGenerator.scala
```diff
@@ -42,7 +42,14 @@ object GoServiceMainCreateHandlerGenerator {
     )
   }
 
-  private def generateDAOCallBlock(root: ServiceRoot, usesMetrics: Boolean, metricSuffix: Option[String]): String =
+  private def generateDAOCallBlock(root: ServiceRoot, usesMetrics: Boolean, metricSuffix: Option[String]): String = {
```
willjvsmith: Could split this function in two, to have `generateDAOCallBlock` and `generateDAOCallErrorBlock` to match the other handler generators. Though I'm not too fussed

willjvsmith (approved):
LGTM 🚀

src/e2e/resources/simple-temple-expected/simple-temple-test-user/dao/dao.go
jaylees14: In the request handler, we match against the type of error and return a response accordingly:
```go
if err != nil {
	switch err.(type) {
	 case dao.ErrSimpleTempleTestUserNotFound:
	         // Return a 404 
         case dao.ErrDuplicateSimpleTempleTestUser:
	         // Return a 403
         default:
                  // Return a 500
	}
}
```

To do that, I define a new type in `errors.go`: 

```go
type ErrDuplicateSimpleTempleTestUser string
```

This requires an existing type for it to be defined against, and string made the most sense to allow for some additional data in the future (such as where the offending constraint was violated etc.)

The alternative to this is to use an error builder:

```go
return nil, errors.New("Error description")
```

The issue with this is that you can no longer pattern match on the type (since it's always Error), and therefore in the handler we can't distinguish between an internal server error (i.e. a 500) and a duplicate user (i.e a 403). We could avoid this by wrapping the error with an associated code of some description, then matching on this code and returning appropriately, but that felt like a bit of over engineering for this small use case, and would increase the complexity of the handler.

jaylees14: Aaaand, scratch all of that. `pq` exposes a "Detail" of the error which I didn't see before, which means I can implement the "additional data in the future" part of what I just said now.

The error is now:

```
Key (bar)=(123) already exists.
```

PR #392, opened 2020-04-26
[Go-gen] Refactor Go handler generation code to prepare for structs

c-harding:
- Generalise a lot of handler functions from ServiceRoot to AttributesRoot
- Use the new `structName` method (returns the name of the struct block, or blank for services) for calculating suffixes, instead of a local `hookSuffix` function
- Remove `enumeratingByCreator`, as it can be calculated from the block, as was already being done in several places
- Remove `usesMetrics` when it can exactly be inferred from the presence/absence of the option `metricSuffix`

N.b. this has no effect on the code output, which will be coming in the next PR.


jaylees14 (approved):
Awesome 😎

Lewiky (approved):
Nice one

PR #393, opened 2020-04-27
[go-gen] Add ParentID to scan call

jaylees14:
Add `ParentID` to the call for scanning into structs.

As per the [Go `sql` documentation](https://golang.org/pkg/database/sql/#Rows.Scan): 

> Scan copies the columns in the current row into the values pointed at by dest. The number of values in dest must be the same as the number of columns in Rows.


c-harding (approved):
ty, my bad

willjvsmith (approved):
LGTM 🚀

PR #394, opened 2020-04-27
[go-gen] Include parentID with DAO input

jaylees14:
Include parentID with DAO input, so it can be stored


Lewiky (approved):
Cool! Do the SQL queries need to be updated too?

c-harding (approved):

c-harding (approved):

jaylees14:
> Cool! Do the SQL queries need to be updated too?

Nah that was already done

jaylees14:
> > Cool! Do the SQL queries need to be updated too?
> 
> Nah that was already done

Saying that, it was done for CREATE but not UPDATE

PR #396, opened 2020-04-27
[metrics] Give unique ID to struct rows

jaylees14:
Every panel we generate in Grafana needs to have a unique ID.

In the current implementation, we were assigning multiple panels the same ID, because we used the index. Rather, we need to use the index, offset by the number of panels we have already generated.


c-harding (approved):
lgtm

willjvsmith (approved):
LGTM 🚀

PR #397, opened 2020-04-27
[openapi-gen] Add struct support to OpenAPI

jaylees14:
Right then... add struct support to OpenAPI.

This reuses as much of the old code as possible, but making it work for structs and services using inheritance.

It also fixes two issue: 
- the ID field in the URL was a number, not `uuid`
- the list endpoint didn't have a 401 error.

I've also noticed that 401 errors are added even when a project doesn't use auth, so I'll rectify that in a follow up PR.


c-harding (approved):
lgtm

src/main/scala/temple/generate/target/openapi/OpenAPIGenerator.scala
```diff
@@ -37,18 +38,28 @@ private class OpenAPIGenerator private (
...
-      description = s"ID of the $lowerName to perform operations on",
-    )
-    paths.getOrElseUpdate(url, Path.Mutable(parameters = Seq(parameter))).handlers
+    val parentIDParameter = when(url.contains("{parent_id}")) {
+      Parameter(
+        InPath,
+        name = "parent_id",
+        required = Some(true),
+        schema = OpenAPISimpleType("string", "uuid"),
+        description = s"ID of the parent entity to perform operations on",
```
c-harding: not sure about this description

c-harding (approved):
looks good, just a description that seems a little odd

src/main/scala/temple/generate/target/openapi/OpenAPIGenerator.scala
jaylees14: What would you suggest?

Lewiky: "ID of the parent entity of the entity to perform operations on"?

Lewiky (approved):
Surprisingly small diff! nice one

PR #398, opened 2020-04-28
[project-test] Add readable and writable to temple test

jaylees14:
Adds readable and writable tests to `temple test`. Everything seemed to work, so very happy 😄

### DIFF
--- a/src/main/scala/temple/test/internal/CRUDServiceTest.scala
+++ b/src/main/scala/temple/test/internal/CRUDServiceTest.scala
@@ -11,6 +11,7 @@ import temple.builder.project.ProjectBuilder
 import temple.generate.CRUD
 import temple.generate.CRUD.CRUD
 import temple.utils.StringUtils.kebabCase
+import temple.builder.project.ProjectConfig.{getDefaultReadable, getDefaultWritable}
 
 class CRUDServiceTest(
   serviceName: String,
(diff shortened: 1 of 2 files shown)


src/main/scala/temple/test/internal/CRUDServiceTest.scala
```diff
@@ -16,6 +20,56 @@ class CRUDServiceTest(
   usesAuth: Boolean,
 ) extends ServiceTest(serviceName, baseURL, usesAuth) {
 
+  private def serviceReadable: Metadata.Readable =
+    service.lookupLocalMetadata[Metadata.Readable].getOrElse {
+      if (usesAuth) Metadata.Readable.This else Metadata.Readable.All
```
c-harding: doesn’t ProjectConfig have something for this?

src/main/scala/temple/test/internal/CRUDServiceTest.scala
```diff
@@ -16,6 +20,56 @@ class CRUDServiceTest(
...
+    }
+
+  private def serviceWritable: Metadata.Writable =
+    service.lookupLocalMetadata[Metadata.Writable].getOrElse {
+      if (usesAuth) Metadata.Writable.This else Metadata.Writable.All
```
c-harding: ditto

c-harding (approved):
very nice

src/main/scala/temple/test/internal/CRUDServiceTest.scala
jaylees14: Ah so it does! Thanks

Lewiky (approved):
LGTM

PR #401, opened 2020-04-28
[openapi-gen] Remove 401 error code if auth isn't used

jaylees14:
We don't need to include a 401 error unless there is a security scheme being used (which is indicative of an auth block in the project) - so.. let's do that!


c-harding (approved):
Very nice

Lewiky (approved):
Nice one

PR #403, opened 2020-04-28
[question-asker] Avoid throwing errors with empty response

### LINKED ISSUE #292: When asked for a go module name, pressing enter causes crash
If you don't type anything for the go module name, press enter to continue, a `sbt.TrapExitSecurityException` is thrown and the application crashes.

jaylees14:
Fix #292.
Also make a nicer error if the input contains whitespace, removing the stack trace.


Lewiky (approved):
LGTMM

c-harding (approved):
lovely

PR #405, opened 2020-04-29
Release/1.0.0

jaylees14:
Release of version 1.0.0

This includes:
- Full Go microservice generation, supporting Create, Read, Update & Delete endpoints, as well as listing stored entries
  - These are currently backed up with a Postgres DB, but we've designed the interface to make it easy to switch this out in the future
  - We've also added support for metrics using Prometheus and Grafana
- Built in authentication using JWTs
  - This allows us to do custom access control on resources, restricting access to only the entity that created it
- Automatic deployment of the project using Kubernetes or Docker Compose, with an API Gateway in place load balancing 
- Automatic generation of OpenAPI schemas, allowing you to generate clients for the API in a single command
- An automatic testing framework for generated projects
  - This spins up the infrastructure locally, makes some sample requests and validates that the responses are what you would expect
- Full suite of documentation (available at https://templeeight.github.io/temple-docs/)
- Full support for custom business logic via hooks
  - This allows our users to define custom validation, add additional database calls etc. to the pre-generated endpoints
- It also allows them to add to or remove from the pre-generated endpoints
  - We also support regeneration of code, without removing any changes they made to their custom logic
  - This means additional fields and services can be added/removed without losing changes they've made
- Examples of how the project could be used to build Uber / Airbnb / Deliveroo etc. (available here:  https://github.com/TempleEight/temple/tree/develop/examples)


Lewiky (approved):

README.md
```diff
@@ -1,3 +1,31 @@
-# temple
+# Temple
 
-[![codecov](https://codecov.io/gh/TempleEight/temple/branch/develop/graph/badge.svg)](https://codecov.io/gh/TempleEight/temple)
\ No newline at end of file
+[![codecov](https://codecov.io/gh/TempleEight/temple/branch/develop/graph/badge.svg)](https://codecov.io/gh/TempleEight/temple)
+![](https://img.shields.io/travis/com/templeeight/temple)
```
c-harding, line 4: Should we get these to show the master version now (and make the landing page the master branch)?

docker/configure_go.py
```diff
@@ -0,0 +1,12 @@
+import json
+import os
+
+data = json.loads(os.environ['v_src'])
```
c-harding, line 4: How did Python sneak into this?

examples/deliveroo.temple
```diff
@@ -0,0 +1,87 @@
+// Deliveroo is a food delivery service with riders, delivering dishes from restaurants to users.
+Deliveroo: project {
+    #language(go);
```
xsanda, line 3: This file is more 4-indented, others are 2-indented: is this desirable?

c-harding (approved):
Wow, we did a lot. I’d say this is first-class work

PR #409, opened 2020-04-29
[kube] Allow the user to define a custom REG_URL

jaylees14:
Based on TempleEight/temple-docs#30, let the user define their own env var for for the registry, rather than editing the generated files.


src/e2e/resources/simple-temple-expected/push-image.sh
```diff
@@ -1,8 +1,8 @@
 #!/bin/sh
-REGISTRY_URL="localhost:5000"
+REG_URL="${REG_URL:-localhost:5000}"
```
Lewiky, line 2: Bloody fancy syntax bash magic that

Lewiky (approved):
Perf

willjvsmith (approved):

c-harding (approved):
bash is magic

PR #413, opened 2020-05-01
[readme-gen] Add struct support to README

jaylees14:
Oh wow I love inheritance ❤️ 

Have some structs!

Also fixes a bug where restrictions on byte lengths were printed as optionals 🙊

### DIFF
--- a/src/main/scala/temple/generate/docs/DocumentationGenerator.scala
+++ b/src/main/scala/temple/generate/docs/DocumentationGenerator.scala
@@ -70,7 +70,7 @@ object DocumentationGenerator {
     }
   }
 
-  private def generateServiceDocs(
+  private def generateDocs(
     builder: MarkupDocBuilder,
     name: String,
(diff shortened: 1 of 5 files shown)


src/main/scala/temple/generate/docs/DocumentationGenerator.scala
```diff
@@ -73,23 +73,24 @@ object DocumentationGenerator {
   private def generateServiceDocs(
```
c-harding: Service?

c-harding (approved):
lvgtm

src/main/scala/temple/generate/docs/DocumentationGenerator.scala
jaylees14: Ah good catch

willjvsmith (approved):
LGTM 🚀 

I think a contents at the top of the file, linking to the docs for each endpoint might be good

PR #414, opened 2020-05-01
Readme proj structure

jaylees14:
On top of #413, so diff will shrink.

Adds project structure to top of README, as well as a link back to us


willjvsmith (approved):
LGTM 🚀 

Definitely think this was worth it, it would be nice if all projects people wrote had a structure description like this in their READMEs

Lewiky (approved):
LGTM

PR #422, opened 2020-05-03
Struct no auth

jaylees14:
Struct create handler would redeclare err when auth block wasn't there, but doesn't need to.


src/test/scala/temple/testfiles/structWithoutAuth.temple
```diff
@@ -0,0 +1,7 @@
+Example: project {}
+
+Brendon: service {
```
Lewiky, line 3: Nice

Lewiky (approved):
LGTM

c-harding (approved):
Nice, my bad, well done

